PR #77, opened 2015-12-08
Fix fallback images for the hero image

jonpacker:
This is an imperfect fix for browsers that don't support `<picture>`. On Safari, it will fall back to `srcset`, which will pick the best image to use depending on the resolution of the browser when the page loads. On IE, it will always show `picture-b`, but it will scale it to 100% width. On Chrome and Firefox it should work as before.

Screenshot of the original layout glitch:

skjermbilde 2015-12-08 kl 13 58 31

### DIFF
--- a/_sass/_components.welcome.scss
+++ b/_sass/_components.welcome.scss
@@ -5,6 +5,10 @@
 *   Site intro and top boxes on the front page
 ------------------------------------------------------- */
 
+picture img {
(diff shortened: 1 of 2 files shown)


robinsk:
Awesome! Thanks for the pull requests, really appreciated :chart_with_upwards_trend: 

@elisabethirgens: You good with this?

PR #78, opened 2015-12-08
Fix proportional SVG scaling in IE

jonpacker:
Everyone's favourite browser, IE, gets confused and assumes that it can't proportionally scale SVGs that have a `width` and `height` property on their `<svg>` element. This PR removes those, so that SVGs render properly.

Screenshot of the glitch before fix:

skjermbilde 2015-12-08 kl 15 52 28

After:

skjermbilde 2015-12-08 kl 15 52 55


robinsk:
Agains, thanks for contributing. @elisabethirgens: I'll leave this in your capable hands as well.

PR #85, opened 2015-12-29
Improve svg icons

elisabethirgens:
The API icons have now been combined as multiple `<symbol>` elements in one single file: **icons.svg** — and then referenced in the HTML with `<use>` and their respective ID.

(Went back and forth on doing the same with logo and illustrations, but decided in the end that they were better of as separate SVG-files for now — so we don’t need to include those at the top of all HTML-files.)

But hey: 218 additions and 407 deletions. 🎉


robinsk:
Niiiice. :shipit:

PR #86, opened 2016-01-05
Changed text for PP API / pickupPointType

chege:
A change has been made that enables the user to filter on pickup point type in Denmark and Finland. Therefore the API text that describes this is updated.


bringdev:
[trello.com/c/hl6q7LAf](https://trello.com/c/hl6q7LAf)

chege:
@Piyushchandra1987 ?

andreaja:
Isn't that available for all countries?

PR #89, opened 2016-02-16
Edi pages

MariusLindberg:
First steps on translating and fixing the EDI part of the developer pages. Text has been translated to English, the provider list is more pretty and details what Nordic country the different providers are available in.

There are more work to be done layout wise but this is hopefully a start.


edi/index.html
```diff
@@ -19,78 +19,117 @@
...
+            <tr>
+            <tr>
+              <td style="text-align: left"><a href="http://www.centiro.com"><h3 class="epsilon u-mb-">Centiro</h3></a></td>
+              <td style="text-align: left"><a href="http://www.centiro.com"><h3 class="epsilon u-mb-">Centiro</h3></a></td>
+              <td style="text-align: left"><a href="http://www.centiro.com"><h3 class="epsilon u-mb-">Centiro</h3></a></td>
+            </tr>
+             <tr>
+              <td style="text-align: left"><a href="https://www.descartes.com"><h3 class="epsilon u-mb-">Descartes</h3></a></td>
+              <td style="text-align: left"><a href="https://www.descartes.com/"><h3 class="epsilon u-mb-">Descartes</h3></a></td>
+              <td style="text-align: left"><a href="https://www.descartes.com/"><h3 class="epsilon u-mb-">Descartes</h3></a></td>
```
robinsk: Why are there three triplets here and for Centiro?

edi/index.html
```diff
@@ -19,78 +19,117 @@
...
+              <th style="text-align: left">Norway</th>
+              <th style="text-align: left">Sweden</th>
+              <th style="text-align: left">Denmark</th>
+            </tr>
+            </thead>
+            <tbody>
+            <tr>
+              <td style="text-align: left"><a href="http://www.consignor.no/"><h3 class="epsilon u-mb-">Consignor</h3></a></td>
+              <td style="text-align: left"><a href="http://www.consignor.se/"><h3 class="epsilon u-mb-">Consignor</h3></a></td>
+              <td style="text-align: left"><a href="http://www.consignor.dk/"><h3 class="epsilon u-mb-">Consignor</h3></a></td>
```
robinsk: Maybe we should just link to consignor.com?

edi/index.html
robinsk: Ah now I see, there's a table head per country. Well, if Consignor is the odd one out, I suggest we just link to the .com version or do something like `Consignor - NO, SE, DK` and link the country codes.

Introducing tables kills the responsiveness, so it doesn't look good on mobile. We should strive for not using tables if we don't have to.

MariusLindberg: image

Works in all nordic countries. URL is apparently the same independent of country. (commit was done by @VidarSten) See image.

edi/index.html
MariusLindberg: @VidarSten you should probably comment on this since you were the one who updated the list.

PR #108, opened 2017-01-17
Add graph mosaic of metrics in bring

kaaveland:
Add a sample of various data that we collect and visualize in the web-team at bring.


AnshikaKoul:
👍

PR #113, opened 2017-02-09
Describe dimension and volume to be optional for cargo

poojajha29:


api/booking/api.raml
```diff
@@ -208,7 +208,7 @@ documentation:
              example: 5
              required: true
         volumeInDm3:
-             description: Specify volume in cubic decimeter
+             description: Specify volume in cubic decimeter. This is optional when dimension is specified in the request for cargo.
              type: decimal
```
ashishveer: example ?

PR #136, opened 2017-06-12
using unix milliseconds since epoch format for shipping date in booking json req

Piyushchandra1987:
normal timestamp in yyy-mm-ddTHH:mm:ss works for xml request but for json request it is not parsed correctly resulting in some issues


mkushe:
This looks OK. But we also need to see how to accept the date, time and timezone in JSON request too.

robinsk:
What, I've used the `yyyy-mm-ddTHH:mm:ss` format in JSON, doesn't it work anymore? That's one of the nice things about the API.

It's very close to the JSON date format (`new Date().toJSON(); // "2017-06-13T07:27:22.731Z"`), so not supporting that in the JSON requests seems super weird.

PR #141, opened 2017-07-21
Remove images and styles for a banner we don’t use

elisabethirgens:
Spring cleaning! (I think we originally had some plans for a banner on each page, but then we changed the layout and focused on documentation and the navigation sidebar.)


imprashast:
looks-good-to-me-ship-it

PR #147, opened 2017-09-01
Document Mailbox Packet (Pakke i postkassen)

robinsk:
This major change adds documentation for the new Mailbox Packet product using RAML 1.0


api/order/to-mailbox/api.raml
```diff
@@ -0,0 +1,541 @@
...
+      postalCode:
+        type: string
+        description: Sender's postal code.
+      postalPlace:
+        type: string
+        description: The city/place for the postal code.
+      email:
+        type: string
+        pattern: ".+@.+"
+        description: Sender's email address that for the order that was placed.
```
farazh: Extra `that`

PR #156, opened 2017-09-27
Setting up jenkins

kjarnet:
Set up build and release to Test and QA from Jenkins with Pipeline and [workflowLibs](https://github.com/bring/workflowLibs/blob/master/vars/javaBuildReleaseDeploy.groovy).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,9 @@
+javaBuildReleaseDeploy {
+    appName = 'developer-site'
+    slackChannel = '#web-dev'
+    buildEnvExtra = ['BUNDLE_PATH=vendor/bundle']
+    def rbenvPrefix = 'source ~/.rbenvrc &&'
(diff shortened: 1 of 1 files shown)


kjarnet:
https://trello.com/c/6Ock4l6n

kjarnet:
Huzzah! Seems to work after I renamed the temporary build directory (for bundle ruby libraries) to start with underscore, so Jekyll doesn't try to parse them 😄

kaaveland:
Looks ok, but I'm wondering if maybe it should deploy to production as well? 🤔 

The "risk" with the current setup is that there's no artifact in the case of developer-site, so it needs to be built for every deploy. That means that the developer who deploys to production from their machine are going to need to make a local build, meaning again, that they'll deploy something different than what jenkins deployed. In the average case, that's fine, but I think there will be occasions when they'll deploy without having pulled the most recent master or forgetting to build first or things like that.

What do you think?

kjarnet:
That's a good point...

PR #157, opened 2017-10-02
Provide frontpage entrance to e-commerce section

elisabethirgens:
This change will reprioritize the three entrances with illustration and headline. Moving e-commerce up to get more focus, and bumping the initial sort intoduction page about Why use our APIs? to a list item in the sitemenu.

### DIFF
--- a/index.html
+++ b/index.html
@@ -16,17 +16,10 @@ layout: redesign
           <h2 class="welcome__tagline">Tools to help you integrate services from Bring with your&nbsp;systems.</h2>
           <ul class="layout welcome__list">
             <li class="layout__item palm-one-whole lap-and-up-one-third welcome__item">
-              <a href="{{ site.baseurl }}/api/introduction/" class="welcome__link">
-                <img src="{{ site.baseurl }}/img/illustration01.svg" alt="">
-                <h2>What is this…?!</h2>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -35,6 +28,13 @@ <h2 class="welcome__tagline">Tools to help you integrate services from Bring wit
                 <h2>EDI documentation</h2>
                 <p class="u-mb0">If you're an EDI vendor, this is where you will find our implementation guides. (Sorry, only in Norwegian at the moment.)</p>
               </a>
+            </li><!--
+          --><li class="layout__item palm-one-whole lap-and-up-one-third welcome__item">
```
andreaja: I don't understand this comment, can you explain it to me?

andreaja:
lgtm, 🐑 🇮🇹 

I noted with some amusement the weakness of numbering images 😂

index.html
elisabethirgens: Yes! This is known as [Fighting the Space Between Inline Block Elements](https://css-tricks.com/fighting-the-space-between-inline-block-elements/) 💥

* elements that are `inline-block` will have space between them
* in this case the visual gap doesn’t matter, but it won’t math nicely
* so to make room for 3 x 33.3% we need to remove the space in the markup
* and imo commenting it out like this is the most readable option

andreaja: TIL! 🎓

PR #197, opened 2018-02-09
Add new job ad

arnemart:


bringdev:
[trello.com/c/tWJtJqcS](https://trello.com/c/tWJtJqcS)

jobs/index.html
```diff
@@ -7,7 +7,7 @@ <h2>Ledige stillinger på webteamet</h2>
 
 <ul>
   <li>
-    <h3><a href="{{ site.baseurl }}/jobs/systemutvikler/">Systemutvikler web</a></h3>
+    <h3><a href="{{ site.baseurl }}/jobs/senior-systemutvikler-web/">Senior systemutvikler web</a></h3>
```
elisabethirgens: The specified `senior` has had a tendency to come and go the past years… I suggest we keep the original URL so it’s always the same.

arnemart: 👍

PR #202, opened 2018-02-12
Add notes on why upgrade

pankajagrawal16:


bringdev:
[trello.com/c/XTUTMDE4](https://trello.com/c/XTUTMDE4)

jonasac:
Should i proof read this, or will we do that later?

pankajagrawal16:
I think we have postponed proof reading till we have a draft

PR #205, opened 2018-02-20
Add http status codes with SG context

satyamagarwal:


bringdev:
[trello.com/c/zlmGxoVT](https://trello.com/c/zlmGxoVT)

PR #209, opened 2018-02-27
Update todo with text update. For now we dont have any page to point …

pankajagrawal16:
For now we dont have any page to point the users. Will be coming soon


bringdev:
[trello.com/c/gYKpUabh](https://trello.com/c/gYKpUabh)

PR #213, opened 2018-03-05
Xsd and wsdl of new version to hosted as version 20 dm a mgq go

pankajagrawal16:


bringdev:
[trello.com/c/DmAMgqGo](https://trello.com/c/DmAMgqGo)

PR #215, opened 2018-03-06
Add text about trace messages

chege:


bringdev:
[trello.com/c/gTKLa4AW](https://trello.com/c/gTKLa4AW)

PR #217, opened 2018-03-06
Changing 72 to 120 days for storing labels

best-ankitagarwal:
Increase store time for shipments in two-step Mailbox API from 72 to 120 hours. if booking is not done within 120 hours of label generation, then it will be deleted by the system in 120 hours now


bringdev:
[trello.com/c/oB6yYiFf](https://trello.com/c/oB6yYiFf)

PR #224, opened 2018-03-18
Go live messages and link updates to SG docs

pankajagrawal16:
image
Thats how it will look on main page and also on documentation of old sg version


pankajagrawal16:
Not it looks like 
image

PR #228, opened 2018-03-26
API terms and conditions.

pankajagrawal16:


_layouts/redesign.html
```diff
@@ -42,6 +42,7 @@ <h1 class="siteheader__tagline">developer</h1>
         <li><a href="https://twitter.com/bringdeveloper" title="Our Twitter account">@bringdeveloper</a></li>
         <li><a href="{{ site.baseurl }}/blog/" title="Bring Developer Blog">Blog</a></li>
         <li><a href="{{ site.baseurl }}/jobs/">Jobs</a></li>
+        <li><a href="https://api.bring.com/shippingguide/user/terms.pdf">Terms & Conditions</a></li>
```
anirbansingharoy: May be we can move 1 layer down under site footer. 
image
I have seen some web sites like google put their terms over there although not sure if these is applicable

pankajagrawal16: I think we can use it like this.

PR #246, opened 2018-06-01
Don’t translate the name Endre to Change

elisabethirgens:
While “Head of Division Change” sounds like a rad title, it makes more
sense to protect this element from translation in translation tools and
we can do that with the `translate` attribute. \o/

### DIFF
--- a/jobs/index.html
+++ b/jobs/index.html
@@ -34,7 +34,7 @@ Teamene har ansvar for hele stacken fra brukerflatene ned til databasen. Du får
 <p>
   Søknader behandles fortløpende. <strong><a href="mailto:arne.moen@bring.com?cc=endre.midtgard.meckelborg@bring.com" class="linkish">Send CV på e-post</a></strong> til Arne og Endre.<br>
   Avdelingsleder Arne Moen, telefon 902&nbsp;02&nbsp;239<br>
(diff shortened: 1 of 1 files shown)


kaaveland:
I'm now going to be calling him that forever 🤣

jobs/index.html
```diff
@@ -34,7 +34,7 @@ <h3>Klar for en prat for finne ut mer? ✏️</h3>
 <p>
   Søknader behandles fortløpende. <strong><a href="mailto:arne.moen@bring.com?cc=endre.midtgard.meckelborg@bring.com" class="linkish">Send CV på e-post</a></strong> til Arne og Endre.<br>
   Avdelingsleder Arne Moen, telefon 902&nbsp;02&nbsp;239<br>
-  Fagsjef Endre Midtgård Meckelborg, telefon 930&nbsp;14&nbsp;504</p>
+  Fagsjef <span translate="no">Endre</span> Midtgård Meckelborg, telefon 930&nbsp;14&nbsp;504</p>
```
kaaveland: Maybe the whole name should be in this `span`? I bet there's some language or tool where his name becomes "Change Middle Farm Meckel Castle"

elisabethirgens: Oh right, that probably makes sense when we’re first whipping out a `<span>`

pankajagrawal16: Oh wow.. i learn something new

PR #256, opened 2018-07-10
Add error code to disallow non-freeze and dangerous goods together

priyamittal:
For API booking for npb-cargo products


bringdev:
[trello.com/c/60lVASLk](https://trello.com/c/60lVASLk)

PR #263, opened 2018-08-20
put general terms condition as a part of bring developer site so that…

manishatgithub:
put general terms condition as a part of bring developer site so that any other application can refer to it. don't think should be a be part of shipping guide.


bringdev:
[trello.com/c/z7scbRBj](https://trello.com/c/z7scbRBj)

manishatgithub:
I can't see this link in customer signup 🤔

manishatgithub:
cleaning would be different task, can you approve if its ok.

pankajagrawal16:
As long as you promise to do the cleanup in SG and update the link here https://www.mybring.com/signup/register/user , it sounds reasonable thing to do.

pankajagrawal16:
Also why is content of https://api.bring.com/shippingguide/user/terms.pdf looks different than what we have here ? is that supposed to change ?

manishatgithub:
bussines has came with new content which need to be updated.

manishatgithub:
Also what is the entry point for https://www.mybring.com/signup/register/user I can't see it on the landing page of signup ?

pankajagrawal16:
Its linked from dev site only as its supposed to be used by API users only.

PR #264, opened 2018-08-22
updating the devsite link for Saturday Delivery

AnshikaKoul:


bringdev:
[trello.com/c/ojmPPaKb](https://trello.com/c/ojmPPaKb)

api/products/index.html
```diff
@@ -531,7 +531,7 @@ <h3 id="additional-services" class="anchored">Additional Services</h3>
               </tr>
 
               <tr>
-                  <td><a target="_blank" href="{{ base_url }}/shippingguide/v2/products/price.xml?weight=1500&amp;frompostalcode=7600&amp;topostalcode=1407&amp;fromcountry=NO&amp;tocountry=NO&amp;product=EKSPRESS09&amp;additionalservice=LORDAGSUTKJORING">LORDAGSUTKJORING</a>
+                  <td><a target="_blank" href="{{ base_url }}/shippingguide/v2/products/price.xml?date=2018-08-24&amp;weight=1500&amp;frompostalcode=7600&amp;topostalcode=1407&amp;fromcountry=NO&amp;tocountry=NO&amp;product=EKSPRESS09&amp;additionalservice=LORDAGSUTKJORING">LORDAGSUTKJORING</a>
```
pankajagrawal16: is not this message wrong `<Message>Additional service "LORDAGSUTKJORING" is not applicable for product EKSPRESS09</Message>` ? as in why is date a factor on VAS 🤔 and then should the message be more explicit ?

AnshikaKoul: ok, ll try explaining it here. so this checks for these two things. 

https://github.com/bring/fraktguide/blob/master/fraktguide-provider-pakker/src/main/java/no/bring/fraktguide/pakker/tilleggstjenester/PakkeAdditionalServiceCalculator.java#L96-L125

*if (lordagsutkjoringsDate == null)* sets this trace message : OMITTING_LORDAGSUTKJORING_ADDITIONAL_SERVICE

*if (!price.isPresent())* sets this trace message : 
LORDAGSUTKJORING_NOT_APPLICABLE


then we get the additional service with these *internal* trace messages. 

which then is resolved here: https://github.com/bring/fraktguide/blob/master/fraktguide-web/src/main/java/no/bring/logistics/fraktguide/service/FraktguideService.java#L106-L128

hope this makes sense. Or we can sit down and I can walk you through this 🙇

AnshikaKoul:
@pankajagrawal16 : how about this? 😁

PR #268, opened 2018-09-13
Add disqus support on blog pages

pankajagrawal16:
image

### DIFF
--- a/_posts/2018-09-05-ship-shape.md
+++ b/_posts/2018-09-05-ship-shape.md
@@ -1,5 +1,6 @@
 ---
 layout: post
+disqus: true
 title: Shipshape Shipping Guide
 date: 2018-09-05 16:00:00 01:00
 category: refactoring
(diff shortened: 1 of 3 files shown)


kaaveland:
This was removed on purpose at some point in the past, so don't go and merge this without a discussion :-)

pankajagrawal16:
@kaaveland Do you remember why ? I thought it will be good idea to see what people have to say about our blogs , if at all some one wants to say anything

elisabethirgens:
I’m not a fan of comment sections. In my experience, it rarely adds value, not for me as a reader and not for me as a writer. There’s a risk of it either being total silence, or just negativity. I would hesitate to write posts if there is a comment section below my post. 

But that said; I am a huge fan of others making their own experiences! ✨ 

@pankajagrawal16 and me had a discussion IRL, and perhaps there’s a way we opt in to add it on some posts, when the author wants to request questions and feedback.

_posts/2018-09-05-ship-shape.md
```diff
@@ -1,5 +1,6 @@
 ---
 layout: post
+disqus: true
```
elisabethirgens: Nice! ✨

PR #277, opened 2018-10-25
Ask for summer internship applications

kaaveland:


jobs/index.html
```diff
@@ -36,6 +36,12 @@ <h3>Klar for en prat for finne ut mer? ✏️</h3>
   Avdelingsleder Arne Moen, telefon 902&nbsp;02&nbsp;239<br>
   Fagsjef <span translate="no">Endre Midtgård Meckelborg</span>, telefon 930&nbsp;14&nbsp;504</p>
 
+<h3>Sommerjobb?</h3>
```
elisabethirgens: How about an ID so we could point people directly to
👉 [developer.bring.com/jobs#sommerjobb](https://developer.bring.com/jobs#sommerjobb) ?

kaaveland: Will do, that's a good idea :-)

elisabethirgens: You can do this directly on the heading element!
`<h3 id="sommerjobb">Sommerjobb?<h3>`

kaaveland: You saw nothing 🙈

PR #281, opened 2018-12-21
Update OTM request with restriction of max 500 supported package numbers

mkushe:


bringdev:
[trello.com/c/CggEfGNu](https://trello.com/c/CggEfGNu)

PR #288, opened 2019-01-16
New way requesting vas for booking api 4 e dz zmx a

cyberq:


bringdev:
[trello.com/c/4EDzZmxA](https://trello.com/c/4EDzZmxA)

api/booking/api.raml
```diff
@@ -19,6 +19,38 @@ documentation:
 - title: Encoding
   content: |
     Booking endpoints will assume all requests are UTF-8 encoded
+      
+- title: A new way of requesting additional services
```
pankajagrawal16: image


I think we should make bullet points and say why they should use only new way.

Ex: All new VASes will be added only on new way., Old way will be decommissioned in some time etc. More flexible etc to make changes when new way is added.

api/booking/api.raml
```diff
@@ -19,6 +19,38 @@ documentation:
...
+            <mobile>12345</mobile>
+          </additionalService>
+          <additionalService>
+            <id>DANGEROUS_GOODS</id>
+            <mobile>777777</mobile>
+            <unNumber>ABC</unNumber>
+          </additionalService>
+       </additionalServices>
+
+       The old way of requesting additional servies:
```
pankajagrawal16: Is there a way to show there two separately in markup and old one with a strike through or something.

pankajagrawal16: Also may be this section should be updated as well. http://developer.test.bring.com/api/booking/#book-shipments

cyberq: can strike through the heading

PR #324, opened 2019-02-25
Remove customs elements from booking api doc

tomgag:
To prevent new clients from using these, as this will be revamped.

### DIFF
--- a/api/booking/api.raml
+++ b/api/booking/api.raml
@@ -212,12 +212,11 @@ documentation:
             description: Specify address line 2. This parameter can be 35 character long.
             type: string
         customsDeclarationType:
-            description: Declaration for customs clarance usage
+            description: _This field is deprecated._ Declaration for customs clarance usage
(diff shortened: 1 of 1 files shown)


AnshikaKoul:
should we also remove them from the examples request/responses?

AnshikaKoul:
or what dag suggested.. that sounds reasonable as well! 

and if we do this, @tomgag we will then need to ask about the deadline. Maria needs to have a dialogue with business about the existing flow. how it works in production. et al.

tomgag:
I agree @dagtornes, marking it as deprecated makes it easier to reference when migrating to new tag. I would have preferred that if I was consuming this service.

tomgag:
Ah, it is not required in the XSD, only in the documentation. I'll remove that.

api/booking/api.raml
```diff
@@ -212,12 +212,11 @@ documentation:
             description: Specify address line 2. This parameter can be 35 character long.
             type: string
         customsDeclarationType:
-            description: Declaration for customs clarance usage
+            description: _This field is deprecated._ Declaration for customs clarance usage
```
mkushe: should we annotate using `@Deprecated`?

PR #329, opened 2019-03-06
Fixing typo

tomgag:


tomgag:
I added goodsDescription, but it does not show. I think the reason is that an element with this name already exists on a higher level in json/xml. Since we flatten the structure here, there is a conflict. Not sure what to do about that one.

PR #355, opened 2019-06-11
Updating booking validation error codes

tomgag:


api/booking/examples/errors.json
```diff
@@ -405,6 +405,46 @@
       "message": "Delivery indoor is available in Sweden and Denmark only.",
       "suffixable": false
     },
+    {
+      "code": "BOOK_VALIDATION-008",
+      "message": "Illegal argument.",
```
AnshikaKoul: this is not customs specific? 🤔

AnshikaKoul:
aha so were there more error codes missing from devsite? 👀like not just the ones we introduced in customs?

tomgag:
Yes, and I only vent through BOOK_VALIDATION* Backlog item to autogenerate these?

api/booking/examples/errors.json
tomgag: no. used by ApiService

PR #368, opened 2019-07-08
Developer site update details invoice endpoint to and f av p6 qfsq

imashishranjan:


bringdev:
[trello.com/c/AvP6Qfsq](https://trello.com/c/AvP6Qfsq)

api/invoice/api.raml
```diff
@@ -52,3 +52,14 @@ documentation:
 
   get:
     description: List invoice numbers
+    queryParameters:
+      fromDate:
+        description: Start date for fetching invoices in format dd.mm.yyyy.
```
nygron: maybe specify what is the default if it is not provided? for both two and from

imashishranjan: Updated

PR #369, opened 2019-07-10
Updated correct url reports -> report

imashishranjan:


bringdev:
[trello.com/c/NaAvXMcT](https://trello.com/c/NaAvXMcT)

PR #374, opened 2019-08-06
Append new FI Posti pickup point type

mkushe:


bringdev:
[trello.com/c/47NAdQzu](https://trello.com/c/47NAdQzu)

PR #402, opened 2019-09-18
Fix wrong year

endremm:


endremm:
Accepting this myself, since it's just a small year correction.

PR #431, opened 2019-11-28
Document return to party tag for pip services ugqc4a64

mkushe:


bringdev:
[trello.com/c/UGQC4a64](https://trello.com/c/UGQC4a64)

PR #434, opened 2019-12-13
Add invoicesource to specified invoice report response

mbrameshkumar:
Updating revision history about this change

http://developer.qa.bring.com/api/revision-history/

### DIFF
--- a/api/reports/api.raml
+++ b/api/reports/api.raml
@@ -39,6 +39,16 @@ documentation:
     5. Download report
     6. Get list of invoice numbers for given customer or group
 
+- title:
+  content: |
+    #### List of available invoice sources
+      - AMPHORA_BCI
+      - AMPHORA_DOMESTIC
+      - PARCEL_DOMESTIC
+      - SYSPED
+      - TIMPEX
+      - AMPHORA_SEA
+
 - title: Authentication
   content: |
     The Reports API requires authentication for all its endpoints. See the [getting started guide on authentication](/api/#authentication) if you're not sure what this means.
@@ -208,8 +218,11 @@ documentation:
     responses:
       200:
         displayName: Report
-        description: Examples show different statuses.
+        description: |
+          Examples show the MetaData included in successful(200) response for specified invoice report vs other reports.
         body:
+          application/xml+invoice:
+            example: !include examples/empty_invoice_report.xml
           application/xml:
             example: !include examples/empty_report.xml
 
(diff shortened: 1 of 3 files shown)


bringdev:
[trello.com/c/i5rFj0G6](https://trello.com/c/i5rFj0G6)

anirbansingharoy:
May be say something about the list of values supported there ?

mbrameshkumar:
> May be say something about the list of values supported there ?

I think that is more detail for revision history. Customer who wants to know more will ask more if they are interested in that value 🤔

anirbansingharoy:
May be not in the Change History page but somewhere in the report documentation it should have  include this information .

api/reports/api.raml
```diff
@@ -39,6 +39,16 @@ documentation:
...
 
+- title:
+  content: |
+    #### List of available invoice sources
+      - AMPHORA_BCI
+      - AMPHORA_DOMESTIC
+      - PARCEL_DOMESTIC
+      - SYSPED
+      - TIMPEX
+      - AMPHORA_SEA
```
nygron: Think we need a little more information. 
Something in the way of:

For specified Invoice report 3 different types of report can be returned depending on the invoice type.
These are.
Cargo identified by 
 - xxxxx xxxx
Parcels  identified by
 - xxx xxxxx 
And the third 
-- xx xxx xxx

mbrameshkumar: @nygron Updated the text with more details.

api/reports/api.raml
```diff
@@ -39,6 +39,26 @@ documentation:
...
+    Cargo International and Sysped :
+
+      - AMPHORA_BCI
+      - SYSPED
+
+    Parcels :
+      - PARCEL_DOMESTIC
+
+    Frigo :
+      - TIMPEX
```
anirbansingharoy: Well as of now the Value it will support is TIMPEX_FR but we have a card separately to fix it . I think we can prioritise  the up next and proceed with TIMPEX now

https://trello.com/c/SBo6JTOE/829-cleanup-feature-toggle-and-related-stuff-added-for-specified-invoice-frigo-report

api/reports/api.raml
```diff
@@ -39,6 +39,26 @@ documentation:
     5. Download report
     6. Get list of invoice numbers for given customer or group
 
+- title:
+  content: |
+    #### List of available invoice sources
+    For specified invoice report 4 different types of report can be returned depending on the invoice type. Those are
```
anirbansingharoy: May be Say "invoice source" instead of "invoice type"

mbrameshkumar:
I am merging this PR and can make change further if required in separate PR since I see one approval 🙂

PR #454, opened 2020-02-25
Update developer site for invoice api

mbrameshkumar:
This PR addresses the following

- Convert the invoice api documentation style to use RAML v1.0
- Structure changes due to migration
- Update the supported customer numbers
- Sample response


bringdev:
[trello.com/c/yAG2vPAV](https://trello.com/c/yAG2vPAV)

api/invoice/api.raml
```diff
@@ -1,66 +1,183 @@
...
+        example: 25.01.2020
+    responses:
+      200:
+        displayName: Successful response
+        description: |
+          The shipment is invoiced based on invoice source and issuer. Here are a list of supported invoice issuers that can be returned.
+
+          | Invoice issuer | Description |
+          |:-------|:--------|
+          | `BRING_CARGO_AS` | `Bring Cargo AS` |
```
anirbansingharoy: I liked the idea here to document this but I am wondering if its needed. These organization we don't specifically show anywhere . Its used in the react app to show invoice source.

api/invoice/api.raml
```diff
@@ -1,66 +1,183 @@
...
+          | `BRING_EIENDOM_HAUGESUND_AS` | `Bring Eiendom Haugesund AS` |
+          | `POSTEN_NORGE_AS` | `Posten Norge AS` |
+          | `POSTEN_NORGE_AS_FOR_CARGO_SEA` | `Posten Norge AS` |
+          | `OTHER` | `Other` |
+
+          Here are a list of invoice types that can be returned.
+
+          | Invoice Source |
+          |:-------|
+          | `INVOICE` |
```
anirbansingharoy: I think the values we show here is "Invoice" , "Credit Note"

anirbansingharoy: I never seen any invoices with "Deposit" and "Debit Note" . Let me check with BKL if they seen any data for these types else we can remove it all together

api/invoice/api.raml
mbrameshkumar: I also feel the same. we can remove it if it is used technically. Earlier I didn't know where it is used.

api/invoice/api.raml
mbrameshkumar: Let me know the outcome. I will make the changes

PR #476, opened 2020-05-04
Add documentation about what customer number to be used in report invoice api

anirbansingharoy:
image

### DIFF
--- a/api/reports/api.raml
+++ b/api/reports/api.raml
@@ -350,6 +350,9 @@ documentation:
   displayName: List invoice numbers
   description: |
     Gets the list of invoice numbers for a customer id or a customer group. By default it shows last 65 days of invoices, with the possibilty to access 365 days data by providing from date and to date.
+
+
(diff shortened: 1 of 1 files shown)


bringdev:
[trello.com/c/wMhLtEiA](https://trello.com/c/wMhLtEiA)

api/reports/api.raml
```diff
@@ -350,6 +350,9 @@ documentation:
   displayName: List invoice numbers
   description: |
     Gets the list of invoice numbers for a customer id or a customer group. By default it shows last 65 days of invoices, with the possibilty to access 365 days data by providing from date and to date.
+
+
+    We support Main Customer Number and Parcel International Customer Number in this API. If you are willing to generate Specified Invoice report, please use mentioned customer number in "List available reports for a customer" endpoint to findout corresponding Report Type.
```
nygron, line 355: "If you are willing" sounds like a challenge. Maybe "If you want to" instead.

nygron:
I'm not confident that this will stop users from generating a specified invoice using the wrong definition. Maybe we need some information in the reports section also? But for sure this will help.   I feel it would have been nice to have a "what has changed since last section" to the api But we are not there yet.

PR #483, opened 2020-05-19
Updated the developer site for CORS support in postalCode api

Surbhi93hub:
Updated the developer site for CORS support in postalCode api

### DIFF
--- a/api/postal-code/api.raml
+++ b/api/postal-code/api.raml
@@ -68,6 +68,11 @@ documentation:
       ### 2. Add authentication headers
       A description of the headers can be found in [Getting Started / Authentication](/api/#authentication).
 
+- title: Support CORS
+  content: |
+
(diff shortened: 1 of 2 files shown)


bringdev:
[trello.com/c/HEPODJ35](https://trello.com/c/HEPODJ35)

api/postal-code/api.raml
```diff
@@ -68,6 +68,11 @@ documentation:
       ### 2. Add authentication headers
       A description of the headers can be found in [Getting Started / Authentication](/api/#authentication).
 
+- title: Support CORS
+  content: |
+
```
mkushe, line 73: should we specify how we support it? as in only in case of authentication

Surbhi93hub: done

PR #485, opened 2020-05-19
Updated dev site with CORS support for authenticated postalCode API

Surbhi93hub:
Updated dev site with CORS support for authenticated postalCode API

### DIFF
--- a/api/postal-code/api.raml
+++ b/api/postal-code/api.raml
@@ -68,6 +68,11 @@ documentation:
       ### 2. Add authentication headers
       A description of the headers can be found in [Getting Started / Authentication](/api/#authentication).
 
+- title: Support CORS
+  content: |
+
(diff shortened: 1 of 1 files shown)


bringdev:
[trello.com/c/HEPODJ35](https://trello.com/c/HEPODJ35)

api/postal-code/api.raml
```diff
@@ -68,6 +68,11 @@ documentation:
       ### 2. Add authentication headers
       A description of the headers can be found in [Getting Started / Authentication](/api/#authentication).
 
+- title: Support CORS
+  content: |
+
+      We support CORS for authenticated postalCode API.
```
mkushe, line 74: may be .. like this => CORS headers will be added to response only for authenticated requests

mkushe: so we explicitly mentioned about authentication

PR #497, opened 2020-07-07
update limit parameter in Get webhooks API documentation

richas2:
image


bringdev:
[trello.com/c/aRunaukR](https://trello.com/c/aRunaukR)

richas2:
> @richas2 Please check if this details needs to be updated in revision history otherwise looks ok

Oh yeah, I'll add it there also. Thanks 🙂

PR #501, opened 2020-07-16
Document non standard courier in api 

Glorina1601:


bringdev:
[trello.com/c/0aR2drL0](https://trello.com/c/0aR2drL0)

api/booking/api.raml
```diff
@@ -497,6 +497,11 @@ documentation:
             type: string
             enum: [SALE_OF_GOODS, RETURNED_GOODS, GIFT, COMMERCIAL_SAMPLE, DOCUMENTS, OTHER]
             required: true
+        nonStandardCourierProductCode:
+            description: Specify the product code for customize courier services. This field is mandatory when request is for Non Standard Courier.
+            type: string
+            example: CDA
```
mkushe, line 503: is there any enum values for this? if yes then specify all the values in here also add that enum in booking xsd

Glorina1601: No, it's a string and it's already there in booking xsd.
```<xs:element name="nonStandardCourierProductCode" type="xs:string" minOccurs="0"/>```

PR #540, opened 2020-10-15
Endpoint area adjustments

halvorsanden:
Adjusts a couple of the endpoint area header levels.
Changes the style of the additional examples toggler and gives them full width.
Removes header classes that override the set font-size.


bringdev:
[trello.com/c/syZEAakj](https://trello.com/c/syZEAakj)

ckaasi (approved):

PR #542, opened 2020-10-15
Consolidate API authentication info

halvorsanden:
Mostly consolidation of API authentication info – adding it as a template piece could have been possible but would have required more work and a few conditionals – not worth it, if things are changing anyway both content- and tech-wise.
Found some visual details to fix here as well 🥳

### DIFF
--- a/api/booking/api.raml
+++ b/api/booking/api.raml
@@ -21,7 +21,7 @@ documentation:
 
 - title: Authentication
   content: |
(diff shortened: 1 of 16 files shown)


bringdev:
[trello.com/c/vb9y9yae](https://trello.com/c/vb9y9yae)

ckaasi (approved):

PR #543, opened 2020-10-16
Nudge header and front page

halvorsanden:
Adjusting the front header, footer and front page to look and behave a bit better and to not take up too much space. The main links were pushed way down on regular screens. Also, links are a bit more obvious when hovering.

It’s just supposed to be a bit better without noticing what has been done – redesign will happen soon anyway.

Oh, the irony that someone called the template "redesign" five years ago 🙊

## before
devsite front pre

## after
devsite front post


bringdev:
[trello.com/c/vb9y9yae](https://trello.com/c/vb9y9yae)

ckaasi (approved):
💅

PR #550, opened 2020-11-02
Improve ETQ documentation

mkushe:
* Removed a param mentioned in REST which was actually needed only for SOAP
* Add new section in SOAP


uksolanki (approved):
👍

PR #557, opened 2020-11-20
Url updates

spadger1066:
Added CreateShipmentRequest and CreateShipmentResponse types to raml
Update to the template to allow multiple types on Request body to facilitate CreateShipmentRequest


_includes/raml_type.html
```diff
@@ -28,6 +28,7 @@ <h4 class="anchored" id="type-{{ typeDefinition.name | slugify }}">{{ typeDefini
             <th>Type</th>
             <th>Description</th>
           </thead>
+          <tr><td>{{ include.type }}</td><td><td></td></tr>
```
endremm, line 31: Double start td tags.

_includes/raml_type_row.html
```diff
@@ -73,6 +73,7 @@
   </td>
 </tr>
 
+<!-- propType is never set to 'object' -->
```
endremm, line 76: If so, should we fix the code?

spadger1066: Yes, there are a number of fixes but I'd like to discuss with you first.

_includes/raml_type.html
spadger1066: Column one only. Neater ways of doing it but it's still a work in progress. This commit is to allow others to review what's been done without my having to share my screen all the time. Will discuss on Monday.

PR #572, opened 2021-01-12
Set up Hugo

halvorsanden:
This sets up Hugo, major templates and some content. Hugo will replace Jekyll.

Keep working in the current structure, not on the content inside the Hugo folder. It’s a work in progress and does not publish anything to production before it is ready. The content in the Hugo folder (APIs, EDI, etc.) is temporary and will be re-copied and overwritten before going live. With one exception being the blog posts: Make new ones in the current structure, they will be copied, but any edits made to older posts must also be made in the Hugo structure, they will no be re-copied.


bringdev:
[trello.com/c/zbLA3czL](https://trello.com/c/zbLA3czL)

PR #582, opened 2021-01-20
Hugo: Env. variable related changes

halvorsanden:
Adds the ability to build with environment variables.
They are used for the Disqus comments and api testing links, I’ve also added env indicator to the header, similar to what we have in Mybring.
This led to some changes in the CSS for the header, which also allowed for adding a menu there, which again led to some adjustments to the footer menu.
This again led to some widening of the wrapper class, which led to the API/EDI pages getting wider, so I replaced the menu and content classes using flexbox and made some improvements to the spacing and borders there (which we will revisit). Which again led to some changes in the height of the page, so the overall API/EDI layout now always forces the footer to the bottom.

I also discovered that the generated files had a lot of whitespaces, and added `-` to the Go template tags except for those that were inline already or not supposed to be inline. https://gohugo.io/templates/introduction/#whitespace
This cut most of the generated files in half, for instance, Booking went from ~16,000 to ~8,000 lines.


bringdev:
[trello.com/c/20eqWelu](https://trello.com/c/20eqWelu)

hugo/README.md
```diff
@@ -8,20 +8,31 @@ Our developer site uses Node, [Hugo](https://gohugo.io/) and
 [RAML 1.0](https://raml.org/). Hugo is a static site generator, and RAML
 is a a YAML-based modeling language for APIs.
 
-There is currently no CSS processing in place. We are just using a copy of the old CSS. This will change.
+We use CSS with some light post processing.
+
+We use Disqus for comments. Disabled by default, it can be activated in the api section using "comments: true".
```
ckaasi, line 13: 👍

ckaasi (approved):
👍

PR #584, opened 2021-01-21
Use Hugo Disqus

halvorsanden:
- Use Hugo’s built-in Disqus functionality, a simpler setup and gives more control in the frontmatter (pages can share the same comments, like in EDI)
- Remove comments option from blog frontmatter, there were two articles that had it and there were zero comments, plus we have not enabled it in the templates now.
- Update comment info in readme


bringdev:
[trello.com/c/20eqWelu](https://trello.com/c/20eqWelu)

ckaasi (approved):
👍

PR #610, opened 2021-02-03
Make revised services subpage of services

halvorsanden:
The index file needs a low dash in the filename to allow for subpages.


halvorsanden:
Ran locally and verified.

PR #626, opened 2021-02-09
Connect mybring design system and remove old css

ckaasi:
- Connect Mybring designsystem
- Replace classes that Mybring designsystem provides
- Add namespace `dev-` to custom css
- Separate and rename custom css 
- Delete unused classes
- Update all templates with new classes

frontpage
gettingstarted
shippingguide
localhost_1313_blog_
localhost_1313_blog_sane-react-config-with-webpack_
localhost_1313_jobs_
localhost_1313_api_pickup-point_


bringdev:
[trello.com/c/FaMopF7l](https://trello.com/c/FaMopF7l)

content/api/_index.html
```diff
@@ -55,10 +55,10 @@ <h3>API status</h3>
   </p>
 </section>
 
-<section class="api__section" id="">
-  <h2 class="anchored" id="authentication">Authentication</h2>
+<section class="dev-documentation__section" id="">
+  <h2 class="anchored dev-documentation__heading" id="authentication">Authentication</h2>
 
-  <p>
+  <p >
```
halvorsanden, line 61: Is there an extra space here?

ckaasi: Ops! Should be fixed now

content/api/_index.html
```diff
@@ -9,18 +9,18 @@
 disqus_identifier: https-developer-bring-com-api
 ---
 
-<h1 class="api__title">Getting started</h1>
-<section class="api__section">
-  <h2 class="anchored" id="introduction">Introduction</h2>
+<h1 class="dev-documentation__title mbl">Getting started</h1>
+<section class="dev-documentation__section">
+  <h2 class="anchored dev-documentation__heading" id="introduction">Introduction</h2>
```
halvorsanden, line 14: heading and subheading for h2 and h3 can be difficult names when we get to h4. We don’t have to make classes for h4–h6, but we should have a naming convention that can be expanded.

It should also work for the things that are inside markdown, where we are unable to apply classes. Which goes for both API and blog. I’m wondering if we could achieve the same with specificity instead of having to apply classes. Like having a class called for instance "dev" in the template, pretty far up, possibly at the body or main element. Then modifying the h elements with `.dev h2` and so on. We can also do a split of `.documentation h2` and `.blog h2`. if needed (although I think h elements should be pretty much the same all over).

content/api/ecommerce/index.html
```diff
@@ -31,20 +22,18 @@ <h1 class="api__title nowrap">E-Commerce</h1>
     represents a part of a suggested order process and contains a list of
     suggested features for that step in the process.
   </p>
-  <br />
-  <br />
 
-  <h3>
+  <h2 class="dev-documentation__subheading">
```
halvorsanden, line 26: Nice h-level fixing!

content/api/services/_index.html
```diff
@@ -1305,7 +1305,7 @@ <h4>Return Services</h4>
     </tbody>
   </table>
 
-  <sup>
+  <sup class="mbl db">
```
halvorsanden, line 1308: Not your mistake, but `sup` is probably not the right element here, can be `p` or `small` element with `text-note` class.

css/base.css
```diff
@@ -0,0 +1,78 @@
+html {
+  font-size: 16px;
```
halvorsanden, line 2: font-size in px overrides the user’s settings, especially this one, this can just be 100 %

But this should be set by the mds stylesheet anyway, since that has resets and base stuff – unless that messes up things too much.

css/blog.css
```diff
@@ -0,0 +1,162 @@
...
+  cursor: default;
+}
+.dev-blog__paginator__button--disabled:hover {
+  border-color: inherit;
+}
+.dev-blog p code {
+  background-color: #ebebeb;
+  padding: 0 0.2ch;
+}
+.dev-blog p + h2 {
```
halvorsanden, line 112: I like this, and we can probably make blog and doc the same on these header settings. Ref previous header comment.

css/table.css
```diff
@@ -0,0 +1,70 @@
+table {
```
halvorsanden, line 1: I think this is the best solution, having a separate table styling without classes. We have the mb-table class, but that won’t work if the tables are in markup etc.

layouts/partials/head.html
```diff
@@ -9,6 +9,11 @@
     {{- with .Title -}}{{ if not ( eq . $.Site.Title) }}{{ . }} –
     {{ end }}{{- end -}}{{ .Site.Title }}
   </title>
+  <link
+    rel="stylesheet"
+    type="text/css"
+    href="{{ "https://www.mybring.com/design-system/css/main.min.css?ts=" | relURL }}{{ now.Unix }}"
```
halvorsanden, line 15: relURL won’t have an effect on this I think. But it is possible to make this somewhat environment conscious, so it fetches qa when running locally, and the other envs when built for those. Did something similar with the baseapiurl in the shortcode folder.

halvorsanden (approved):
A couple of comments, things we can discuss further also, but nothing stopping this from merging when you are ready. Really good to see Inuit go and more familiar things coming in +  better-structured documents 👍

PR #674, opened 2021-03-10
Indicate manned/unmanned pickup point types in Norway

mariadlis:


torstein-a (approved):
👍

content/api/pickup-point/api.raml
```diff
@@ -19,13 +19,12 @@ documentation:
       #### Distinct PickupPointType
 
       Norway :
-        - Type 4  : Post i Butikk
-        - Type 19 : Posten pakkeutlevering
-        - Type 21 : Postkontor
-        - Type 27 : Pakkeautomat
```
AnshikaKoul, line 25: pakkeautomat needs to be removed?

mariadlis: Yes doing a couple of more changes, was just waiting for a confirmation from business. Seems like I made 2 PRs..

PR #686, opened 2021-03-17
Add HTTP to the headers column

tomgag:
image

image

### DIFF
--- a/content/api/_index.html
+++ b/content/api/_index.html
@@ -105,7 +105,7 @@ disqus_identifier: https-developer-bring-com-api
   <table>
     <thead>
       <tr>
-        <th>Header</th>
+        <th>HTTP header</th>
         <th>Example</th>
         <th>Comment</th>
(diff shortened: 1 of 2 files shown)


halvorsanden:
Does this mean that "header name" here should also be "HTTP header"?
Screenshot 2021-03-17 at 11 29 08

tomgag:
> Does this mean that "header name" here should also be "HTTP header"?

I guess we could keep it `HTTP header name` if that makes more sense. I just wanted HTTP in there, as we've been getting some questions regarding this.

halvorsanden:
> I guess we could keep it `HTTP header name` if that makes more sense. I just wanted HTTP in there, as we've been getting some questions regarding this, and I noticed there's no mention of HTTP here

Only `HTTP header` for both seems fine. The other one is in `docs.html` if you want to change it.

halvorsanden (approved):

PR #689, opened 2021-03-18
Sidemenu as fixed and drawer

halvorsanden:
- Fixes the sidemenu when above 1000 px and hides it in a drawer when below.
- Makes header a bit smaller, since it will be fixed on the docs pages.
- Prevents the header form becoming too tall in IE11 – there are a couple of bugs here and there, so we will do a separate pass to fix some of that, but I don’t think we will spend much time trying to fix minor issues for such a low and dwindling number of users.
- Removes and updates some of the JS from the Jekyll version.

Screen Shot 2021-03-19 at 10 45 55
Screen Shot 2021-03-19 at 10 45 40
Screenshot_2021-03-18 Getting started with Bring APIs – Bring Developer(1)
Screenshot_2021-03-18 Getting started with Bring APIs – Bring Developer


bringdev:
[trello.com/c/zZwd9xk2](https://trello.com/c/zZwd9xk2)

ckaasi (approved):
👍

PR #693, opened 2021-03-18
Fix prominent layout issues in ie11

halvorsanden:
Some quick fixes to get the pages to display correctly in IE11.
There are still a few alignment issues in the docs, but they are too small to spend any time on and would require a lot of extra code. Going below 1024 px width causes more serious issues, but I doubt there are a lot of IE on pads and phones.


bringdev:
[trello.com/c/iH1UoYBV](https://trello.com/c/iH1UoYBV)

ckaasi (approved):

PR #720, opened 2021-04-12
Make checkout guide figs. sticky

halvorsanden:
- Reverse main flex axis, so the descriptions are to the left, but still below the figure when it wraps.
- Make checkout figs sticky, except the one that has two since they pretty much takes up the entire height of the text content.
- Organise table and example links vertically, to give the table more space.


bringdev:

bockelie93 (approved):

PR #743, opened 2021-04-26
Final updates to list of NO pickup points

mariadlis:


mariadlis:
@AnshikaKoul I guess we should make a separate card to remove type 27 and 33 from our code as well?

AnshikaKoul (approved):

PR #744, opened 2021-04-27
Verify existence of VAS before iterating

halvorsanden:
- Stops empty rows from appearing if referencing non-existing VASes
- Stops build from crashing if a service doesn’t have a VAS array


bringdev:

bockelie93 (approved):

PR #746, opened 2021-04-27
Documentation layout adjustments

halvorsanden:
- Makes some pages narrower and some wider
- Slight adjustment of overall layout – tried to avoid using percentages, but in order to get the scrollable menu, flex or grid isn’t an option since both causes double scrollbars on the right side.


bringdev:

ckaasi (approved):

PR #759, opened 2021-05-06
Add parameters and response tables to the Checkout Guide

halvorsanden:
The Checkout Guide needs to refer to parts of the API documentation. To do this, selected data from Booking, Shipping guide 2 and Pickup point APIs are presented on separate pages along with additional hard-coded content. The plan is to fetch everything from the API data files, but putting the new things in there is a task in itself.

These pages are still hidden from the navigation, unless you’re actually on the page.

This is a first working version, a bit fragile if the API changes too much (unlikely) and the templates start to get a bit big and full of conditionals for different page types – so that will require some cleanup when the data and content are being reworked.


bringdev:
[trello.com/c/j9aEhRc0](https://trello.com/c/j9aEhRc0)

ckaasi (approved):

PR #783, opened 2021-05-25
Make scalable fig for ETA

halvorsanden:
Lots of divs and grid to get this to scale – but it works, in IE as well, and it’s a fig and not an interface – so I think it’s OK.

Screenshot 2021-05-25 at 09 33 32
Screenshot 2021-05-25 at 09 33 58


bringdev:
[trello.com/c/n0NQLpBk](https://trello.com/c/n0NQLpBk)

ckaasi (approved):

PR #787, opened 2021-06-01
Update pickup parcel and eta text

halvorsanden:


content/api/checkout-guide/pickup-parcel.html
```diff
@@ -31,7 +31,7 @@
 ---
 
 <section class="dev-docscontent__section">
-  <h2 class="mbm">Pakke til hentested i chekout</h2>
+  <h2 class="mbm">Pakke til hentested in checkout</h2>
```
ckaasi, line 34: Should this text be in Norwegian?

ckaasi (approved):

content/api/checkout-guide/pickup-parcel.html
halvorsanden: Looks a bit weird, but no. The guide uses the Norwegian service names in most places, even though the language is English. I’m thinking there’s room for improvement on that. Some places both are displayed though.

PR #821, opened 2021-07-02
Update preliminary freight economy report availability status

mbrameshkumar:


bringdev:
[trello.com/c/jQbqx9HR](https://trello.com/c/jQbqx9HR)

content/api/reports/_index.html
```diff
@@ -12,10 +12,10 @@
 ---
 
 <div class="message--warn maxw48r pam mbl">
-  <strong>The Preliminary freight calculation report for Parcel Norway customers will be disabled from 21.06.2021</strong>
+  <strong>The Preliminary freight calculation report for Parcel Norway customers were disabled from 21.06.2021</strong>
```
larsaane, line 15: was disabled....

mbrameshkumar: Parcel Norway customers is plural hence i added as `were`..may be I will make it to Parcel Norway customer was disabled from June 21st. What do you think?

larsaane: Of course you are right :)

larsaane: Rewrite to customer was... so it is the same for all the text you changed.

larsaane (approved):
Good

PR #848, opened 2021-08-05
Enums

spadger1066:
Enum inclusions

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3858,9 +3858,9 @@
       }
     },
     "underscore": {
-      "version": "1.11.0",
-      "resolved": "https://registry.npmjs.org/underscore/-/underscore-1.11.0.tgz",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -3858,9 +3858,9 @@
       }
     },
     "underscore": {
-      "version": "1.11.0",
-      "resolved": "https://registry.npmjs.org/underscore/-/underscore-1.11.0.tgz",
-      "integrity": "sha512-xY96SsN3NA461qIRKZ/+qox37YXPtSBswMGfiNptr+wrt6ds4HaMw23TP612fEyGekRE6LNRiLYr/aqbHXNedw==",
+      "version": "1.13.1",
```
endremm, line 3861: If you only have some simple content change, I suggest to skip this file. Then you are sure nothing breaks.

endremm (approved):

PR #904, opened 2021-10-13
Add changelog item: When booking parcel services from Finland, weight…

mdybo:
… and dimensions will be mandatory to set in the request


mariadlis:
@mdybo Look good 👍

PR #942, opened 2021-11-05
Make a structure outline template for the api docs

halvorsanden:
Restructures the API pages into a more coherent order and description as well as using less space in the top:
1. Intro to make it clear what the API does, this is new but heavily based on existing content
2. Common types of information, not all have the same types or the same details
3. API-specific technical information (mostly left untouched except when the intro already covered it)
4. How-to and guides
5. Endpoints

A few of the APIs are left unchanged until their intros are ready.

## Example
Screenshot 2021-12-14 at 15 45 28


bringdev:
[trello.com/c/ISBQjx44](https://trello.com/c/ISBQjx44)

PR #963, opened 2021-11-22
Enable search via algolia

jonasac:


bringdev:
[trello.com/c/U51XyUwA](https://trello.com/c/U51XyUwA)

jonasac:
Screenshot 2021-11-22 at 14 24 25
Screenshot 2021-11-22 at 14 24 32
Screenshot 2021-11-22 at 14 25 06

I'm using menu and the summary part of the search index now, i think we can improve on sizes and maybe position of the box, but lets first find out what should be part of a search hit here. The component/menu breakdown in mds is not that useful here since everything would say api. I am only indexing things where section == api

halvorsanden:
I agree, let’s figure out the refinements.
I think we can exclude revision-history.

Starting to see the problems/mess of having some API documentation as checkout guide subpages, even though temporarily, they will be around for some months still. That’s not a search problem though, it’s a structure problem.

arnemart:
Superseded by https://github.com/bring/developer-site/pull/1275

PR #965, opened 2021-11-24
Service portfolio

halvorsanden:
Changes the service list to a service portfolio. Uses data sources for the booking + shipping guide services table and the reports + invoice services table. This is supposed to make it easier to update for us and easier for the user to filter and get an overview of very large tables.

The VAS tables are also being worked on: https://github.com/bring/developer-site/pull/993 For now, they are just pasted in, hardcoded like today – with some table syntax improvements. The rest is all new template and JS.

When all three tables groups are done, we can look into if some functions can be shared as well as transpiling and/or minifying. 

Screenshot 2021-12-23 at 15 07 55


bringdev:
[trello.com/c/CY3RyncB](https://trello.com/c/CY3RyncB)

PR #1010, opened 2022-01-18
Use same names as in service portfolio families

halvorsanden:


bringdev:
[trello.com/c/UURuVItQ](https://trello.com/c/UURuVItQ)

PR #1013, opened 2022-01-19
Service portfolio syntax adjustments

halvorsanden:
This mainly turns `a` elements that aren’t links into buttons and anchorizes ids used as anchors.

Linking to each one is still buggy as the cutoff is in the way and the items don’t open (and elements above the item that is supposed to be opened becomes unclickable). But that’s a separate task to fix.

### DIFF
--- a/css/vas.css
+++ b/css/vas.css
@@ -2,6 +2,11 @@
     max-height: 24rem;
 }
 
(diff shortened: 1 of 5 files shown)


cdwinsnes (approved):
👍🏻

PR #1014, opened 2022-01-20
Show VAS sections when anchor links clicked

Igor-J86:
- Expand VAS element and scroll to it if anchor link is clicked
- If URL contains anchor link, automatically display the whole VAS table, hide the cut-off, scroll to that VAS element by its ID, and expand that VAS element

### DIFF
--- a/css/documentation.css
+++ b/css/documentation.css
@@ -6,12 +6,12 @@
   height: 100%;
 }
 
-.dev-anchored:target::before {
+/*.dev-anchored:target::before {
   content: "";
   display: block;
   height: 6.5rem;
   margin: -6.5rem 0 0;
-}
+}*/
 
 .dev-docs__page {
   display: flex;
--- a/css/documentation.css
+++ b/css/documentation.css
@@ -6,12 +6,12 @@
   height: 100%;
 }
 
-/*.dev-anchored:target::before {
+.dev-anchored:not(#allVas *):target::before {
   content: "";
   display: block;
   height: 6.5rem;
   margin: -6.5rem 0 0;
-}*/
+}
 
 .dev-docs__page {
   display: flex;
--- a/static/js/servicefiltering-vas.js
+++ b/static/js/servicefiltering-vas.js
@@ -20,6 +20,10 @@ function vasHideCutoffs() {
   vasCutoff.hidden = true
 }
 
+function vasShowCutoffs() {
+  vasCutoff.hidden = false
+}
+
 function clearFilters() {
   vasHideCutoffs()
   const activeBtn = document.querySelector("[data-filterbtn].active")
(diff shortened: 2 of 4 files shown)


bringdev:
[trello.com/c/lLcj6iPh](https://trello.com/c/lLcj6iPh)

css/documentation.css
```diff
@@ -6,12 +6,12 @@
...
 }
 
-.dev-anchored:target::before {
+/*.dev-anchored:target::before {
   content: "";
   display: block;
   height: 6.5rem;
   margin: -6.5rem 0 0;
-}
+}*/
```
halvorsanden, line 14: If I remember correctly, this compensates for the fixed header. Creating an offset so the item that is linked to doesn’t end up behind the header, but just below it. I can see that it creates some problems when it overlays buttons and similar, possibly a negative z-index can help, but haven’t tested it.

Igor-J86: Offset for header is being handled with the new JS logic.

halvorsanden: That’s fine, but it’s still in use in other parts of the site - example: https://mango-cliff-0e97d9203-1014.westeurope.azurestaticapps.net/api/shipping-guide_2/#versioning

Maybe we should split that into a separate class or something …

static/js/servicefiltering-vas.js
```diff
@@ -172,6 +172,52 @@ filterBtns.forEach((filterBtn) => {
   })
 })
 
+function scrollToVasTarget(targetElement){
+  const element = document.getElementById(targetElement)
+  const headerOffset = document.querySelector('.dev-siteheader').offsetHeight + 20
```
halvorsanden, line 177: Should probably reference an id or something not tied to the CSS, lower likelihood of breaking if someone refactors the CSS down the line.

static/js/servicefiltering-vas.js
```diff
@@ -172,6 +172,52 @@ filterBtns.forEach((filterBtn) => {
...
+  const elementPosition = element.getBoundingClientRect().top
+  const offsetPosition = elementPosition + window.pageYOffset - headerOffset
+
+  const vasBtnParent = element.parentElement
+    
+  const vasBtnToggler = vasBtnParent.querySelector('[data-collapse]')
+  vasBtnToggler.classList.remove('dev-collapsible__toggler--collapsed')
+  vasBtnToggler.classList.add('dev-collapsible__toggler--expanded')
+
+  const vasItemContent = vasBtnParent.querySelector('.dev-collapsible__item--collapsed')
```
halvorsanden, line 187: Same thing here about separating CSS and JS concerns.

halvorsanden (approved):
Just a couple of comments, can be merged when fixed 👍

css/documentation.css
Igor-J86: This will be applied to all other places except in the VAS are, by using ":not" selector in CSS.

static/js/servicefiltering-vas.js
Igor-J86: Added data-attributes to those elements, and selecting rather them in JS instead of classes.

static/js/servicefiltering-vas.js
Igor-J86: Added data-attributes to those elements, and selecting rather them in JS instead of classes.

PR #1016, opened 2022-01-25
Publish API update for stopping courier bicycle in norway

AnshikaKoul:


bringdev:
[trello.com/c/OwhvnD6p](https://trello.com/c/OwhvnD6p)

package-lock.json
```diff
@@ -1,5281 +1,8 @@
 {
   "name": "developer-site",
   "version": "2.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
cdwinsnes: Seems like you have an old version of npm locally that will generate a legacy version of the package-lock. Since you have made no changes to the npm dependencies it is no point in commiting this change. This file should only be modified when dependency changes are done.

Since you have an old version of npm, running`npm install`will rewrite the lock file to the old format even though there are no dep changes.

cdwinsnes (approved):

PR #1048, opened 2022-02-16
OAS response area

Igor-J86:


bringdev:
[trello.com/c/NPIJUsZw](https://trello.com/c/NPIJUsZw)

css/documentation.css
```diff
@@ -166,3 +166,22 @@ summary:hover {
   flex-basis:20rem;
 }
 
+.response summary {
+  border:1px solid transparent;
+}
+.response.success-code summary {
```
halvorsanden, line 172: Let’s also add :hover to these, both red and green, but also one for default/gray.
The green one can reuse the colour from the summary elements at the top of the page, we might have some colour defined for the red in the design system buttons page.

layouts/_default/api.html
```diff
@@ -180,6 +187,43 @@ <h3>Response examples</h3>
...
+                schema.classList.add('dn')
+              })
+              
+              showSelectedSchemaOption(this)
+            })
+          }
+
+          function showSelectedSchemaOption(sel) {
+              let opt;
+              for ( let i = 0, len = sel.options.length; i < len; i++ ) {
```
halvorsanden, line 210: Can this be a forEach loop?

layouts/_default/api.html
```diff
@@ -119,26 +121,30 @@ <h2>{{ .summary }}</h2>
...
+                                  {{ $selected := "selected" }}
                                   {{ range $applicationType, $_ := .content }}
-                                    <option value="{{ $applicationType }}"> {{ $applicationType }}</option>
+                                    <option value='schemaType-{{$response}}-{{$endpointId}}-{{$applicationType}}' {{$selected}}>   {{ $applicationType }}
+                                    </option>
+                                    {{$selected = ""}}
                                   {{ end }}
                                 </select>
-                              </h4>
+                              </p>
```
halvorsanden, line 142: I would have used a label element instead of p, and given the select a form__control class. Look at the form element page in the design system. We can always place them inline and adjust the width if we want to.

halvorsanden (approved):
Feel free to merge after considering the comments. 👍
A bit border-rich, but we can look at that section as a whole instead of detail controlling it now.

layouts/_default/api.html
Igor-J86: Done

layouts/_default/api.html
Igor-J86: I will take a look..

Igor-J86: forEach loop is only working for Arrays, Sets, and Maps. This function is called multiple times from another function with forEach loop, and is applying necessary stuff on each single element. So, I think we'll just keep it like this for now, and we can review it later again if needed. :)

PR #1055, opened 2022-02-18
Response area iteration

Igor-J86:


bringdev:
[trello.com/c/NPIJUsZw](https://trello.com/c/NPIJUsZw)

halvorsanden (approved):

layouts/_default/api.html
```diff
@@ -112,13 +112,13 @@ <h2>{{ .summary }}</h2>
...
 
-                        {{- partial "api/params.html" (dict "parameters" .parameters) -}}
+                      {{- partial "api/params.html" (dict "parameters" .parameters) -}}
 
                         <p class="text-title">Responses</p>
 
                         {{ $responseCodeColor := "" }}
                         {{ $successCodes := "200 204" }}
-                        {{ $errorCodes := "400 404 500" }}
+                        {{ $errorCodes := "400 404 429 500" }}
```
halvorsanden, line 121: A thought (maybe something for after Address launch?), maybe these could be a switch with ranges of sorts? If code is between 400 and 600, return error class … and so on.

Igor-J86: Yes, we need to do this in a smarter way.

PR #1056, opened 2022-02-22
OAS examples

halvorsanden:
This is part of a work in progress, but still hidden except for the new API.

It adds support for examples that are either directly tied to a response code or via refs. The examples are read as JSON from the JSON file. The functionality for switching between the examples and subexamples is also in place. The styling is not done at all. The layout needs more work as well.

response.html and responsejson.html will either be merged into one file or have split functionality – keeping both now for reference for each other.

Next up is doing some additional formatting and extracting the JS into separate files and bundling them.

### DIFF
--- a/layouts/_default/api.html
+++ b/layouts/_default/api.html
@@ -173,11 +173,34 @@
                         {{ end }}
                         {{ range $response, $_ := .responses }}
                           <div id="{{ $response }}">
-                            <pre class="chroma"><code>
-{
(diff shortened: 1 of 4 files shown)


bringdev:
[trello.com/c/l9A1QhN8](https://trello.com/c/l9A1QhN8)

Igor-J86 (approved):
Looks good to me :)

PR #1064, opened 2022-02-24
Add example,enum,type where available

Igor-J86:


bringdev:
[trello.com/c/bLGHuc0m](https://trello.com/c/bLGHuc0m)

Igor-J86:
Last state:
Skjermbilde 2022-02-24 kl  13 51 06

layouts/partials/api/params.html
```diff
@@ -1,19 +1,69 @@
...
+
+      {{ if and (reflect.IsSlice $enum) ( gt (len $enum) 0 ) }}
+        <div class="ptxs mbs">
+          {{ if gt (len $enum) 1 }}
+            <p class="text-note mb0">Enum:</p>
+          {{ else }}
+            <p class="text-note mb0">Value:</p>
+          {{ end }}
+          {{ range $enum }}
+            <code class="paxs bg-gray3 mbxs db maxwmaxc">{{.}}</code>
```
halvorsanden, line 52: It seems unnecessary to restyle the code elements when they have (or should have) an almost identical style already. I would have tried changing the styling on the type information instead, maybe not using badge, but possibly `text-note` and `gray`.

Igor-J86: Adjusted in new commit where divs are transformed into description lists, and code elements look better without additional styles applies (except some margin and display class for the enum part).

layouts/partials/api/params.html
```diff
@@ -1,19 +1,69 @@
...
     {{ $paramsHeader = .in }}
-    <h4 class="ttu mb-border bb mbs pbxs mtl"> {{ $paramsHeader }} parameters</h4>
+    <h4 class="ttu mb-border bb mbs pbxs mts"> {{ $paramsHeader }} parameters</h4>
   {{ end }}
   
-  <div class="flex mbm">
-    <div class="flex-1">{{ .name }}
+  <dl class="desc-list flex">
+    <dt>
+      <code class="bg-white">{{ .name }}</code>
```
halvorsanden, line 16: this also need `fwn` or it will get a blurry ghosting effect in regular screens – maybe set the fwn on the dt element for now?
Screenshot 2022-02-24 at 14 17 14

Igor-J86: Yes, done.. :)

layouts/partials/api/params.html
```diff
@@ -1,19 +1,69 @@
...
+          {{ end }}
+        {{ else if eq $key "example" }}
+          {{ $example = delimit (slice $name $value) "=" }}
+        {{ end }}
+      {{ end }}
+
+      {{ if and (reflect.IsSlice $enum) ( gt (len $enum) 0 ) }}
+        <div class="ptxs mbs">
+          {{ if gt (len $enum) 1 }}
+            <p class="text-note mb0">Enum:</p>
```
halvorsanden, line 47: Didn’t think of this before, but `p` inside description list can be problematic. A dt element can have multiple dd elements though. Maybe we should stick to spans or divs here for now? And reevaluate element usage after launch?

Igor-J86: Oh, right.. Didn´t think of it myself... Will make the change to divs for now.. .)

halvorsanden (approved):
Each parameter is now a separate list – it might require some tinkering to get a new dl to start after a header and to get the end tag in the right place. Feel free to merge and make a new task for this after launch, if you don’t want to spend time on it now.

PR #1067, opened 2022-02-25
Request area refinement

halvorsanden:
- Added header to the area – looks like "request" can be a better name than "parameters"
- Made the columns flex wrap
- Made each header type into a `dl` by iterating over header types instead of the parameters alone
- various spacing, border and detailing
- Tried to format as little as possible to avoid a bunch of whitespace changes this time (let’s do a separate PR with a bunch of that instead)
- Still uncertain about some details, let’s discuss further improvements to this after launch

Screenshot 2022-02-24 at 19 09 24

### DIFF
--- a/css/documentation.css
+++ b/css/documentation.css
@@ -206,12 +206,17 @@ summary:hover {
 .schema-type-select {
   height:auto;
 }
-.desc-list {
-  margin-bottom:0;
+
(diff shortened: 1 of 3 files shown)


bringdev:

Igor-J86 (approved):
Looks good! :)

PR #1068, opened 2022-02-25
Format request endpoint type and address

Igor-J86:
Skjermbilde 2022-02-25 kl  09 43 58


bringdev:

halvorsanden:
How about having the method outside the grey area, with a slight spacing between them? I think we could get rid of the border then and use the badges classes for the method (with extra padding).

Then we could make a blue one for POST when we get to that in the other APIs – and we already have a red one for DELETE.

Igor-J86:
Agree.. Changed to:
Skjermbilde 2022-02-25 kl  10 26 10

halvorsanden (approved):
Looking good!

PR #1070, opened 2022-02-25
Spacing and endpoint overview adjustments

halvorsanden:
- Some spacing adjustments to get a better sense of the different layout groups
- Renamed and rearranged the endpoint sections – "overview of endpoints" always felt a bit long and is difficult to find in the menu. Placed "usage" first in the table. More detail work can be done here, maybe we can replace the table. This was done on both the new and the current versions.


bringdev:

Igor-J86 (approved):
Looks good!

PR #1078, opened 2022-03-01
Improve button for expanding responses

Igor-J86:


bringdev:
[trello.com/c/4SntNFBJ](https://trello.com/c/4SntNFBJ)

Igor-J86:
Skjermbilde 2022-03-01 kl  09 47 43

layouts/partials/api/schema.html
```diff
@@ -66,7 +66,7 @@
                     </div>
                   {{- end -}}
                 </dt>
-                <dd>
+                <dd class="paxs plm">
```
halvorsanden, line 69: This causes the following as far as I can see:
Screenshot 2022-03-01 at 09 52 23

I have something for the spacing between dt/dd in the pipeline on the other task.

Igor-J86: It shows like this here...? 
Skjermbilde 2022-03-01 kl  09 56 18

css/documentation.css
```diff
@@ -232,6 +232,12 @@ summary.error-code:hover {
   font-weight: 400;
 }
 
+.param-toggle-btn {
```
halvorsanden, line 235: I would have created separate classes for use in JS and CSS, so one can be changed, renamed or moved without the other one breaking.

layouts/partials/api/schema.html
halvorsanden: Funky difference between Chrome and FF by the looks of it.

css/documentation.css
Igor-J86: Ok, will change the class name and add it to the button as well. :)

layouts/partials/api/schema.html
Igor-J86: But I will remove it and we can handle the spacings here and there in another task.

halvorsanden: The left padding on the dt elements without buttons cause the width of it to increase – not sure why it’s different in the two browsers though. I’d say keep it for now – I’m working on removing some of the padding to balance the different levels anyway.

halvorsanden (approved):

PR #1079, opened 2022-03-01
Improve structure and responsiveness of schema data

halvorsanden:
Makes the output sturcture of the schema follow a logic of:
dl > dt + dd + dd > restarting if nested ♾
With divs where needed
This makes the responsiveness work better
Screenshot 2022-03-01 at 15 55 27

Paddings and margin have also been adjusted to have things align better

The request parameter data has also been sorted to be like the response data – the template has also been simplified and there is room for more, probably not as simple as schema-dd.html, but along those lines.

### DIFF
--- a/css/documentation.css
+++ b/css/documentation.css
@@ -234,10 +234,6 @@ summary.error-code:hover {
   padding: 0;
 }
 
-.desc-list dd code {
-  overflow-wrap: anywhere;
-}
-
 .desc-list dd {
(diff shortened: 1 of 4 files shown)


bringdev:
[trello.com/c/THpDWhn7](https://trello.com/c/THpDWhn7)

Igor-J86 (approved):
🚀

PR #1104, opened 2022-03-10
Fix $name value to .name

best-ankitagarwal:
It is currently displaying wrong property name in the example. as the  range operator on the map is not working in the same order. 

Screenshot 2022-03-10 at 6 35 45 AM

Screenshot 2022-03-10 at 6 48 55 AM


bringdev:
[trello.com/c/Ri2Lav4w](https://trello.com/c/Ri2Lav4w)

PR #1108, opened 2022-03-14
Reports and invoice combo filtering

halvorsanden:
Adds combination filtering to Reports and Invoice table similar to how the booking and shipping guide works.
Keeps track of the selected filters in an object, but has to loop through that object in a given order so the rendering works.
Had to redo/simplify the text input filtering as a result of this.

Will do a separate PR with some formatting – didn’t want to introduce a lot of extra whitespace changes now.


bringdev:
[trello.com/c/1kDgcOmP](https://trello.com/c/1kDgcOmP)

PoshYeti (approved):

PR #1110, opened 2022-03-15
Changed the suggested doc

faizal1234786:
- made changes in the doc by adding Sweden, Finland because these are not using Geonames along with Norway as was in present source code of PostalCodeImporter class in PickUpPoint repo


best-ankitagarwal (approved):
Looks good 👍

PR #1086, opened 2022-03-03
Add return filter option to Booking and Shipping Guide table

halvorsanden:
Reordered and refactored much of the file to implement the possibility to have single button filters and filtersets work side by side, and not just hard coding a return filter.

PR #1111, opened 2022-03-15
Redo BSG return service type filtering

halvorsanden:
Reverts the return filter functionality from https://github.com/bring/developer-site/pull/1086 – most refactoring is kept though, and we’ll do a separate refactor of the entire service portfolio.

Instead of having a return filter, we have added it as service types B2B return and B2C return, and removed C2B.

Data changes made in https://github.com/bring/mybring-service-country-vas-common/pull/163


bringdev:
[trello.com/c/zotChbnJ](https://trello.com/c/zotChbnJ)

Igor-J86 (approved):
🚀

PR #1117, opened 2022-03-18
Refactor service portfolio JS

halvorsanden:
- Lots of renaming of functions and consts, removing the prefixes used when we started out and had no tooling to handle that stuff
- Made some names clearer and similar across the files
- Extracted a few obvious functions for reusability
- Added the no matches UI to the VAS filtering

CSS and template refactor is up next.
There is also a known bug in the RI filter where report filtering shows nothing after toggling the API filters in combination. WIll try to fix that in this project.

### DIFF
--- a/css/documentation.css
+++ b/css/documentation.css
@@ -6,7 +6,7 @@
   height: 100%;
 }
 
-.dev-anchored:not(#allVas *):target::before {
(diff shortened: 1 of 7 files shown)


bringdev:

PoshYeti (approved):

PR #1119, opened 2022-03-22
Use this instead of e.target where possible

halvorsanden:
This means we don’t have to use pointerevent CSS on icons inside queryselected elements to prevent them from preventing the click if the user hits exactly on the icon.


bringdev:

PR #1120, opened 2022-03-22
Put reports and invoice table in a separate file

halvorsanden:
Also renames the data attributes from `data-ritreptype` to `data-ri-reptype` so it’s a bit easier to read when referencing them using `dataset.riReptype` instead of `dataset.rireptype`


bringdev:

PoshYeti (approved):

PR #1121, opened 2022-03-22
Split Booking and Shipping Guide table from service portfolio

halvorsanden:
First PR for splitting all three parts.
This also includes making/moving/fixing some common partial templates as well.


bringdev:

PR #1131, opened 2022-03-29
Update doc for 1736  and 5600 ETA

best-ankitagarwal:


bringdev:
[trello.com/c/372bmKOy](https://trello.com/c/372bmKOy)

AnshikaKoul (approved):

AnshikaKoul:
let's wait @best-ankitagarwal with this, as we should only update it when we move from PH to ETA for all customers and that will not happen till ETA supports polygon stuff.
so ll :white_check_mark: it but we should park it for later and not merge it right now :)

PR #1138, opened 2022-04-05
Remove btn-link--help after extraction to mds

halvorsanden:
Was added in https://github.com/bring/mybring-design-system/pull/653


bringdev:
[trello.com/c/jZGPikUu](https://trello.com/c/jZGPikUu)

PR #1139, opened 2022-04-05
Support response examples without components

Igor-J86:
Add simplified version where it checks for response examples directly in the responses, without looking for examples referenced in the "components" section.

The other version that looks for examples in "components" is still kept, but the file is renamed.


bringdev:
[trello.com/c/aq0W8HJi](https://trello.com/c/aq0W8HJi)

halvorsanden:
Looks like the menu is generated in the staging version. I didn’t think that would happen – looks unrelated to this PR though, but let’s figure it out before merging this.
I also thought it would be possible to reuse more code, but maybe it’s a bit too early to focus on that now – two versions isn’t bad at all, especially if the alternative is a lot of conditionals.

Igor-J86:
Added filter for Tracking API, so the menu is not generated in production/staging (as it is for Pickup Point API).

halvorsanden (approved):
👍

PR #1147, opened 2022-04-11
Deeper nesting of schema response codes

Igor-J86:
Remaining:
- Figure out a way to avoid duplicate IDs where params are parents (rendered as buttons), in combination with the belonging nested list.
- Check if some data is missing from the deep


bringdev:
[trello.com/c/aq0W8HJi](https://trello.com/c/aq0W8HJi)

halvorsanden (approved):

PR #1159, opened 2022-04-25
Show position of page content in left menu

Igor-J86:
See if IntersectionObserver makes sense to use here.

### DIFF
--- a/layouts/partials/api/docs.html
+++ b/layouts/partials/api/docs.html
@@ -333,4 +333,47 @@ if('clipboard' in navigator) {
     }
   }
 }
+
+//Show in left menu which endpoint is in viewport
+const anchoredTitles = document.querySelectorAll('.dev-anchored')
+
+if(anchoredTitles) {
+  const listItems = document.querySelectorAll('[data-list-item]')
+
+  let observerOptions = {
+    root:document.body.main,
+    rootMargin: '-45% -45%',
+    threshold: 0
+  }
+  
+  const ob = new IntersectionObserver(observeFunc,observerOptions)
+
+  anchoredTitles.forEach((el) => {
+    ob.observe(el)
+  })
+
+  function observeFunc(entries) {
+    entries.forEach(entry => {
+      if(entry.isIntersecting) {
+        listItems.forEach(item => {
+          item.classList.remove("active")
+        })
+        let listItem = document.querySelector('[data-list-item="'+entry.target.id+'"]')
+        if(listItem) {
+          listItem.classList.add('active')
+        }
+      }
+    })
+  }
+  listItems.forEach(li => {
+    li.addEventListener('click', function() {
+      setTimeout(function() {
+        listItems.forEach(item => {
+          item.classList.remove("active")
+        })
+        li.classList.add('active')
+      }, 10)
+    })
+  })
+}
 </script>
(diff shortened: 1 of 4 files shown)


bringdev:
[trello.com/c/kw0pYwe6](https://trello.com/c/kw0pYwe6)

layouts/partials/sidemenuitem.html
```diff
@@ -32,16 +33,17 @@
...
           </a>
         </li>
       {{- end -}}
 
       {{- with (index .apiData .identifier) -}}
         {{- range .resources -}}
+          {{- $itemId := .displayName | anchorize -}}
           <li>
-            <a href="#{{ .displayName | urlize }}">
+            <a href="#{{ .displayName | urlize }}" data-list-item="{{$itemId}}">
```
arnemart, line 46: Should maybe be `.displayName | urlize` instead of `$itemid`, items with dashes in them don't work this way

layouts/partials/api/docs.html
```diff
@@ -333,4 +333,47 @@ <h4 class="dev-anchored" id="{{ $responseID }}">
...
+
+  function observeFunc(entries) {
+    entries.forEach(entry => {
+      if(entry.isIntersecting) {
+        listItems.forEach(item => {
+          item.classList.remove("active")
+        })
+        let listItem = document.querySelector('[data-list-item="'+entry.target.id+'"]')
+        if(listItem) {
+          listItem.classList.add('active')
```
arnemart, line 363: Add `listItem.scrollIntoView({ behavior: "smooth", block: "center", inline: "nearest" })` for a fun scroll effect! Also, you can consider moving lines 358-360 inside this if to avoid removing the highlight if there is no new item to highlight.

layouts/partials/sidemenuitem.html
halvorsanden: Alternatively `anchorize` instead of `urlize` if these are anchors. https://gohugo.io/functions/anchorize/

Igor-J86: I used `anchorize`, but it resulted in couple of extra dashes if the item name already contains a dash. Seems to be more "stable" with `urlize`.

layouts/partials/api/docs.html
Igor-J86: Thanks! Done ✓

layouts/partials/sidemenuitem.html
halvorsanden: OK 👍

arnemart (commented):
The code is duplicated, maybe add it to a separate file and include it in both places?

Igor-J86:
> The code is duplicated, maybe add it to a separate file and include it in both places?

We have one JS file that is in use for the new Open API Swagger documentation pages, and for the rest of the pages this file is not included. Therefore the code is currently duplicate since we are in a transition fase, but not all APIs are ready with their Swagger documentations.

static/js/scrollmarker.js
```diff
@@ -0,0 +1 @@
+!function(){const e=document.querySelectorAll(".dev-anchored");if(e.length>0){const t=document.querySelectorAll("[data-list-item]");let o={root:document.body.main,rootMargin:"-45% -45%",threshold:0};const c=new IntersectionObserver((function(e){e.forEach((e=>{if(e.isIntersecting){let o=document.querySelector('[data-list-item="'+e.target.id+'"]');o&&(t.forEach((e=>{e.classList.remove("active")})),o.scrollIntoView({behavior:"smooth",block:"center",inline:"nearest"}),o.classList.add("active"))}}))}),o);e.forEach((e=>{c.observe(e)})),t.forEach((e=>{e.addEventListener("click",(function(){setTimeout((function(){t.forEach((e=>{e.classList.remove("active")})),e.classList.add("active")}),10)}))}))}}();
```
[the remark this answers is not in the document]
Igor-J86: Right, didn't think about that.. Done ✓

PR #1162, opened 2022-04-28
Add 1163 VAS info

pedromacli:
1163 VAS info


bringdev:
[trello.com/c/EjobpIs8](https://trello.com/c/EjobpIs8)

tomgag (approved):
LGTM. Not sure we need full requests for the examples here. But I see that's what we've been doing for other VASes as well.

PR #1184, opened 2022-05-20
Add hidden WIP page for OAS sourced Modify Delivery API

kiangol:
index.md needed a low dash in order to allow for child pages within the same dir.
The new page is hidden in the menu and from crawlers and robots


kiangol:
[Modify-Delivery: go live with Swagger on dev site](https://trello.com/c/Q4qQ1eAZ/2700-modify-delivery-go-live-with-swagger-on-dev-site)

halvorsanden:
Schema check has been added in https://github.com/bring/developer-site/pull/1193 - try rebasing and see if it works as intended.

kiangol:
> Schema check has been added in #1193 - try rebasing and see if it works as intended.

Looks good 👍

halvorsanden:
No request/parameters for POST endpoints?
Either way, I should add a conditional in the template to avoid rendering the header if there is none.

Is it correct that there are no response schemas as well?

kiangol:
> No request/parameters for POST endpoints? Either way, I should add a conditional in the template to avoid rendering the header if there is none.

POST endpoints have shipmentNumber parameter in path, as well as request body. Body also is provided as example. 
> 
> Is it correct that there are no response schemas as well?

No response schemas for now. 

We will have to rewrite/clean up this, as it is should not be necessary to send the same information in both path and body.

halvorsanden (approved):

PR #1185, opened 2022-05-20
Add aria tablist and tabpanel role for examples

Igor-J86:


bringdev:
[trello.com/c/dEIqAcqC](https://trello.com/c/dEIqAcqC)

css/documentation.css
```diff
@@ -177,6 +177,10 @@ summary:hover {
   width: 38%;
 }
 
+button[aria-selected="true"] {
+  background-color: hsla(0, 0%, 0%, 0.12);
+}
+
```
halvorsanden, line 183: Should we add something like this to the design system at https://github.com/bring/mybring-design-system/blob/9c43289dcbd56e984f23fb6ed53fb433edc12513/design-system/content/css/components/buttons.css#L175 and https://github.com/bring/mybring-design-system/blob/9c43289dcbd56e984f23fb6ed53fb433edc12513/design-system/content/css/components/buttons.css#L180 instead?

Igor-J86: I think that makes more sense, so it can be accessible in other apps as well. I will remove it from here and rather add it to MDS CSS.

halvorsanden (approved):

PR #1193, opened 2022-05-27
Check for schema content not just content

halvorsanden:
Makes it possible to have examples without having schemas
Extra special conditional used since go’s break tag doesn’t exist in go template https://github.com/golang/go/issues/20531


bringdev:
[trello.com/c/3OqrOFLh](https://trello.com/c/3OqrOFLh)

Igor-J86 (approved):
🚀

PR #1194, opened 2022-05-30
Defensive header rendering

halvorsanden:
Renders request and response headers only when the sections have actual content.


halvorsanden:
Needed in https://github.com/bring/developer-site/pull/1184

Igor-J86 (approved):

PR #1198, opened 2022-06-07
Updating docs to reflect updated VAS booking codes

torstein-a:
Had to rename example files for 1123 and 1139 for them to work (but not for the installation VASes for some reason?)

One interesting find is that all the examples used the correct booking code, so seems the wrong codes in the lib must be a bug.


anurag331998 (approved):
Looks good 👍

PR #1204, opened 2022-06-09
Set up booking oas page

halvorsanden:
Duplicates the existing page and removes most of its content (just copy and paste it back in again if needed)
The new page is hidden from the menu and has no comment section until it is ready to replace the current page.


chandnaankit (approved):
LGTM

PR #1205, opened 2022-06-09
Simplify OAS publishing

halvorsanden:
The conditionals complicated things and just lead to more questions, the new solutions is as follows:
If you add OAS param, the new documentation will show up locally and in prod.
If you don't want that behaviour, you need to make a new, hidden file with the oas stuff.


bringdev:
[trello.com/c/hRisMxDt](https://trello.com/c/hRisMxDt)

PR #1213, opened 2022-06-20
removing REST GET information from environmental data

dipyayan:
removing REST GET information from environmental data section as part of removal of REST GET operations from SG.


ankit22a (approved):
👍

PR #1221, opened 2022-06-23
Only render $ref examples once

halvorsanden:
- Prevent additional values inside example objects with $ref to render as separate examples
- Use description as name if it exists, instead of the key even though they usually are the same


ankit22a (approved):
Looks Good 👍

PR #1224, opened 2022-06-24
Add XML schema

halvorsanden:
- Renders XML specific schema
- Keeps the root level open by default
- Some visual details are different from the JSON render, but that will be fixed in an upcoming task to improve how that is rendered. 
- The major difference in this new template is that the entire .schema is passed to it before looping. And we check the given type in the data source instead of checking the type of the actual data structure. We trust that an object is an object if it says so – but we’re instead being more defensive about what it might contain. The basic structure is easier to follow:
  - If it’s a ref to a different place in .schema, go find the data for it and pass it in again
  - If it’s an object or an array (with descendants), start a new level and iterate/pass in the descendants one by one.
  - If it’s none of the above, it can just be shown, no need to iterate further.

Note that examples are still only JSON.

### DIFF
--- a/css/documentation.css
+++ b/css/documentation.css
@@ -221,7 +221,7 @@ summary.error-code:hover {
 .schema__sublist {
   border-left-style: dashed;
   border-width: 1px;
-  margin: 0 0.75rem;
+  margin: 0 0.25rem 0 0.75rem;
 }
 
 .desc-list > div {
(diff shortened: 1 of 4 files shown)


bringdev:

Igor-J86 (approved):
Looks good to me..!  👍🏻

PR #1228, opened 2022-07-01
Param and schema details

halvorsanden:
- Put "Schema" as header inside responses
- Use "Media type" as label instead
- Use a p element instead of select when there’s only one media type
- Use slightly bolder text instead of colon for example/enum/default
- Use plural s when there are more than one example
- Make the parameter text regular size, but keeping small size in schema.

I think some of these can be improved further, the example/enum/header looks a bit too important in some cases. Maybe background sectioning could be an option in the next round.

Screenshot 2022-07-01 at 06 53 56
Screenshot 2022-07-01 at 07 08 23

### DIFF
--- a/layouts/_default/api.html
+++ b/layouts/_default/api.html
@@ -151,20 +151,31 @@
 
                             {{- if $hasSchema -}}
                               {{- with .content -}}
+                              {{- $multiSchema := false -}}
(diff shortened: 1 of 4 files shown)


bringdev:

Igor-J86 (approved):
Looks good! 🚀

PR #1231, opened 2022-07-01
Add XML example functionality

halvorsanden:
The XML conversion and display solution is solid. It supports both OAS specified XML schema properties that it combines with the JSON examples as well as pure string examples.

The JS for switching between the different types is quick and dirty, but it works OK. Lots of room for improvement (made a card for it).


bringdev:

Igor-J86 (approved):
Looks really good! 👍🏻
The copy button is present on request examples, but I cannot see any request examples, so I suppose it looks good and works as expected. It's removed from the response examples, but I think that should be fine. It's more important to keep it on the request examples. 🙂

halvorsanden:
Yeah, the copy button was removed from responses – made a reminder to add it back in (if it’s needed).
Several of the APIs have new hidden subpages with the OAS version: Pickup Point, Tracking, Booking.
And OM has request example as well, I think.

PR #1232, opened 2022-07-01
Add guides section with content from pickup point

halvorsanden:
Quickly adding support for this.
Made a task to do it properly, separating the section more from the rest of the content etc.

Screenshot 2022-07-01 at 14 45 17


Igor-J86 (approved):
👍🏻

PR #1235, opened 2022-07-13
improving the pickup point documentation

dipyayan:
The PR has been raised to improve the pickup point documentation.

Previous text : For precise results, we strongly recommend to use the end customer’s complete address when using the API.

Present text : For precise results, we strongly recommend to use the end customer’s complete address (not only their postal code) when using the API.


ankit22a (approved):
Looks good 👍

PR #1247, opened 2022-08-05
Add endpoint method to anchor links

Igor-J86:
Adding endpoint method to anchor links and id's, in order to solve potentially duplicate id's if the endpoint name is the same for different methods (ie. POST, GET, DELETE etc.).


bringdev:
[trello.com/c/1lFT4Zwp](https://trello.com/c/1lFT4Zwp)

halvorsanden (approved):

PR #1248, opened 2022-08-05
Make json template code more like the xml version

halvorsanden:
Not necessarily fewer lines of code. Although it’s possible to eliminate some of the repetition, that will make it less readable. Compared to the XML template which just takes everything and shows it, the JSON version has fewer levels. I guess smaller output requires bigger templates.

The styling is slightly different (similar to the XML now) and I’ve added "array of …" instead of just "array". Arrays that doesn’t have any items are no longer able to be opened (no need to open empty boxes).

### before
Screenshot 2022-08-05 at 16 38 11

### after
Screenshot 2022-08-05 at 16 37 22


bringdev:
[trello.com/c/kzHHZXp4](https://trello.com/c/kzHHZXp4)

Igor-J86 (approved):
🚀

PR #1255, opened 2022-08-19
Support dark mode

Igor-J86:
One way of adding support for dark mode, by setting a data-attribute for dark theme on :root (`html`) and defining other colors for the different parts of the dev-site based on that. A CSS only approach is neat, but since we are transforming custom color properties into regular hex-colors, this seems to be a bit tricky to deal with. 

In this way we are overriding the different parts with colors we already have (plus added a couple of new ones), and can easily set the dark theme if the user has that enabled in the OS settings with JS. We can also then store what the user prefers in the local storage if he/she has clicked/tapped the button for manually setting the theme.

Maybe there's a smarter approach, but instead of saying that `--green-light` should be a gray color, I found it somehow easier to use the colors as they are and rather override the different parts with them. 
There are still some colors that need to be considered, but making this PR so we can do some testing.

Additionally I stumbled across a small thing that was missing in the endpoints urls, and fixed that.


bringdev:
[trello.com/c/NDOODZgD](https://trello.com/c/NDOODZgD)

halvorsanden (approved):

PR #1258, opened 2022-08-23
Add postcss-nesting dependency

Igor-J86:
Adding `postcss-nesting` dependency, so we can utilize this on the CSS files where we can use the nesting method.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
         "postcss-css-variables": "^0.18.0",
         "postcss-import": "^14.1.0",
         "postcss-loader": "^7.0.0",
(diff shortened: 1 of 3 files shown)


bringdev:
[trello.com/c/F1gY1pF2](https://trello.com/c/F1gY1pF2)

halvorsanden:
Remember to add it to the config as well: https://github.com/csstools/postcss-plugins/blob/main/plugins/postcss-nesting/INSTALL.md#webpack
Let me know if you need any help and we can look at it together.

halvorsanden (approved):

PR #1268, opened 2022-09-02
Support rest and soap on the same page for OM

halvorsanden:
- Makes a third layout option where old and new has to live on the same page. I’ve tried to reuse as much as possible without polluting or increasing the complexity of the existing templates.
- In the OM api.raml file, I moved the request body parameters into the properties object, which allowed us to use much of the new template code for this.
- The code is specific to OM, so any other APIs with the same need, will probably have to do some adjustments – this keeps the code small when we don't have to take everything into account.


bringdev:
[trello.com/c/C2gux6xW](https://trello.com/c/C2gux6xW)

Igor-J86 (approved):
Nice work! 🚀

PR #1272, opened 2022-09-08
Make a guides section on the API pages

Igor-J86:
A dedicated "Tips and guides" section, that looks somewhat different from the rest of the content. Usually I would steer away from using a border for outlining the section and rather add a light gray background, but this somehow works best with the light green details elements. When that's said, they could also have another background colour.

Added also an icon to indicate that this section might be worth taking a look at. 

Had to also make a small adjustment on the details elements where the `<p>` had some surrounding padding. I removed that, and added rather a `<div>` with some surrounding padding. When there's a `<ul>` in the details element, it works best this way, since the bullet of the list item doesn't float outside the left aligned content.

Skjermbilde 2022-09-08 kl  12 17 52

Skjermbilde 2022-09-08 kl  12 19 51


bringdev:
[trello.com/c/7JZnFWTs](https://trello.com/c/7JZnFWTs)

Igor-J86:
Directly to that section:
https://mango-cliff-0e97d9203-1272.westeurope.azurestaticapps.net/api/pickup-point/#tips-and-guides

content/api/pickup-point/index.md
```diff
@@ -16,27 +16,30 @@ introduction: |
...
-    Max parcel dimensions for Pakkeboks (parcel lockers in Norway) pickup points can now be used to filter out pickup points based on parcel size on the client side, and thus preventing failed bookings during checkout. JSON field:
-    ```json
-      "maxParcelDimensions": {
-        "length": 44.5,
-        "width": 50.5,
-        "height": 60.0
-      }
-    ```
+- heading: Tips and guides
+- introText: Here you will find relevant tips and guides for the API.
```
halvorsanden, line 20: I think we should be strict with the structure here, so this should go into the template – and only show up if `guides` has content.
I also think we can remove `introText` as it just repeats what `heading` communicates.

layouts/partials/api/oas/tips-and-guides.html
```diff
@@ -0,0 +1,31 @@
...
+      </h2>
+    {{- end -}}
+    {{- with .introText -}}
+      <p>{{ . }}</p>
+    {{- end -}}
+    {{- with .sections -}}
+      <div class="dev-docscontent__info owls">
+        {{- range . -}}
+          <details class="bg-green1 rad-a2px">
+              <summary class="pas plm" aria-label="{{.title}} details dropdown">
```
halvorsanden, line 16: Not necessary to have aria on the summary. The element, content and even state is announced, as far as I can tell.

layouts/partials/api/oas/tips-and-guides.html
```diff
@@ -0,0 +1,31 @@
+<div class="mb-border ba bw1 pam mtl maxw44r">
+  {{- range . -}}
+    {{- with .heading -}}
+      <h2 class="dev-anchored" id="{{ . | urlize }}">
+        {{ . }}
+        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 384 512" class="icon-ui--m light-icon"><path d="M112.1 454.3c0 6.297 1.816 12.44 5.284 17.69l17.14 25.69c5.25 7.875 17.17 14.28 26.64 14.28h61.67c9.438 0 21.36-6.401 26.61-14.28l17.08-25.68c2.938-4.438 5.348-12.37 5.348-17.7L272 415.1h-160L112.1 454.3zM192 0C90.02 .3203 16 82.97 16 175.1c0 44.38 16.44 84.84 43.56 115.8c16.53 18.84 42.34 58.23 52.22 91.45c.0313 .25 .0938 .5166 .125 .7823h160.2c.0313-.2656 .0938-.5166 .125-.7823c9.875-33.22 35.69-72.61 52.22-91.45C351.6 260.8 368 220.4 368 175.1C368 78.8 289.2 .0039 192 0zM288.4 260.1c-15.66 17.85-35.04 46.3-49.05 75.89h-94.61c-14.01-29.59-33.39-58.04-49.04-75.88C75.24 236.8 64 206.1 64 175.1C64 113.3 112.1 48.25 191.1 48C262.6 48 320 105.4 320 175.1C320 206.1 308.8 236.8 288.4 260.1zM176 80C131.9 80 96 115.9 96 160c0 8.844 7.156 16 16 16S128 168.8 128 160c0-26.47 21.53-48 48-48c8.844 0 16-7.148 16-15.99S184.8 80 176 80z"/></svg>
```
halvorsanden, line 6: I like this and generally the way it is pointed out as a special section.

content/api/pickup-point/index.md
Igor-J86: Agree..! Adjusting it now..

layouts/partials/api/oas/tips-and-guides.html
Igor-J86: Right..! Didn't think of that.. Removing it now.. 🙂

layouts/partials/api/oas/tips-and-guides.html
Igor-J86: Yes, the icon becomes somehow an eye catcher. 👀

halvorsanden (approved):
Nice solution! I agree on the border v. background colour decision.

PR #1275, opened 2022-09-12
Implement algolia search

arnemart:
- Make hugo output an `algolia.json` file with all of the relevant data we want to index
- Send this data to algolia on every production build
- Add react component for search to the page

A lot can probably be done to make the searching experience nicer but this is probably a good first step.


halvorsanden (approved):
Let’s get this out as soon as you are ready, and we’ll be prepared for some feedback-based adjustments. Very happy to see this 🚀

static/js/algolia.js.LICENSE.txt
```diff
@@ -0,0 +1,37 @@
+/*!
+  Copyright (c) 2018 Jed Watson.
```
halvorsanden, line 2: Do we need this file commited?

arnemart: Nope

PR #1281, opened 2022-09-19
Use mds mb-disclosure class for summary/details

halvorsanden:
Making some changes based on https://github.com/bring/mybring-design-system/pull/679

### DIFF
--- a/layouts/partials/subscribe.html
+++ b/layouts/partials/subscribe.html
@@ -13,8 +13,8 @@
             <label class="checkbox subfield form__check mbm" for="gdpr_366129"><input type="checkbox" id="gdpr_366129" name="gdpr[366129]" value="Y" class="av-checkbox gdpr">I confirm that Posten Norge AS can send me API update emails</label>
           </fieldset>
         </div>
-        <details class="mbl bg-gray1 rad-2px">
-          <summary class="pas">Privacy policy</summary>
+        <details class="mb-disclosure mbl bg-gray1">
+          <summary>Privacy policy</summary>
           <div class="text-note pas content__gdprLegal">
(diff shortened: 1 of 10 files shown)


bringdev:
[trello.com/c/Px2Qpmyj](https://trello.com/c/Px2Qpmyj)

layouts/partials/subscribe.html
```diff
@@ -13,8 +13,8 @@
             <label class="checkbox subfield form__check mbm" for="gdpr_366129"><input type="checkbox" id="gdpr_366129" name="gdpr[366129]" value="Y" class="av-checkbox gdpr">I confirm that Posten Norge AS can send me API update emails</label>
           </fieldset>
         </div>
-        <details class="mbl bg-gray1 rad-2px">
-          <summary class="pas">Privacy policy</summary>
+        <details class="mb-disclosure mbl bg-gray1">
+          <summary>Privacy policy</summary>
           <div class="text-note pas content__gdprLegal">
```
Igor-J86, line 18: Looks mostly nice, but in dark mode it looks kind of funky. I think maybe the `summary` causes it..? 🙂

halvorsanden: Dark mode updated!

Igor-J86: Nice! There are some small light corners on the "red" responses, while on the "green" they are not shown. It seems like the class `.bg-red1` on the `details` element is not overridden in dark mode, but the "green" above is.
Skjermbilde 2022-09-20 kl  13 49 00

halvorsanden: That’s it! I gotta see an optometrist. 🤓
Good catch! I’ll fix.

PR #1282, opened 2022-09-20
Improve solution for switching between media types

Igor-J86:
Currently the switching between data formats in responses section and examples section are two separate things. With this PR they are melting together, so the switching between data formats (usually JSON or XML) is becoming global. 
If you choose one data format in the responses section, the same format is chosen for all other response sections, but also in all the example sections, and vice versa.

The chosen format is also then stored in local storage, so we can make a (cautious) presumption that the user wants to see the same format across the other API pages or when revisiting the same page, and preselect the preferred data format globally for the user. 🙂

### DIFF
--- a/js/api.js
+++ b/js/api.js
@@ -1,46 +1,4 @@
-// Responses
-const schemaTypeSelect = document.querySelectorAll(".schema-type-select")
-schemaTypeSelect.forEach(changeType)
-
-function changeType(type) {
-  type.addEventListener("change", function () {
-    const selectedOption = this.value
-    const schemaTypeParent = document.querySelector(
-      '[data-schematype-id="' + selectedOption + '"]'
-    ).parentNode
-    const schemaTypes = schemaTypeParent.querySelectorAll(
-      ".schema-type-container"
-    )
-
-    schemaTypes.forEach((schema) => {
-      schema.classList.add("dn")
-    })
-
-    showSelectedSchemaOption(this)
-  })
-}
-
-function showSelectedSchemaOption(sel) {
-  let opt
-  for (let i = 0; i < sel.options.length; i++) {
-    opt = sel.options[i]
-    if (opt.selected === true) {
-      const schemaType = opt.value
-      const schemaTypeContainer = document.querySelector(
-        '[data-schematype-id="' + schemaType + '"]'
-      )
-      schemaTypeContainer.classList.remove("dn")
-      return schemaType
-    }
-  }
-}
-
-function showInitialSchemaType() {
-  schemaTypeSelect.forEach(showSelectedSchemaOption)
-}
-
-showInitialSchemaType()
-
+//Param toggle buttons
 const paramToggleBtn = document.querySelectorAll(".param-toggle-btn")
 
 paramToggleBtn.forEach((btn) => {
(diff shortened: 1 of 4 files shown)


bringdev:
[trello.com/c/jd0OiNGL](https://trello.com/c/jd0OiNGL)

js/api.js
```diff
@@ -112,40 +70,70 @@ responseExBtns.forEach((btn) => {
...
+        format = storedFormat
+      }
+      //Set the selected option based on format
+      if(f.options[i].value === format) {
+        f.options[i].selected = true
       } else {
-        example.hidden = true
+        f.options[i].selected = false
+        //Fallback for json or xml if format don't entirely match with option values
+        if(format.includes("json") && f.options[i].value.includes("json") || format.includes("xml") && f.options[i].value.includes("xml")) {
```
halvorsanden, line 95: Will this be easier to read with parenthesis around each side of the `||`?

js/api.js
```diff
@@ -112,40 +70,70 @@ responseExBtns.forEach((btn) => {
...
+  formatSelects.forEach((f) => {
+    for(let i = 0; i < f.options.length; i++) {
+      //Set the format from local storage (if stored), with fallback to json or xml if something else
+      if(storedFormat && (storedFormat.includes("json") || storedFormat.includes("xml"))) {
+        format = storedFormat
+      } else if(f.options[i].value === storedFormat) {
+        format = storedFormat
+      }
+      //Set the selected option based on format
+      if(f.options[i].value === format) {
```
halvorsanden, line 90: Can lines 90–96 be restructured into:
if (f.options…) {true}
else if (format.includes…) {true}
else {false}

Or will that mess up the logic?

js/api.js
```diff
@@ -112,40 +70,70 @@ responseExBtns.forEach((btn) => {
...
+  
+  dataTypes.forEach((dataType) => {
+    //Display response parameters and examples based on format
+    if(dataType.dataset.type === format) {
+      dataType.hidden = false
+    } else {
+      dataType.hidden = true
+      //Fallback for json or xml if format don't entirely match with example data-type
+      if(format.includes("json") && dataType.dataset.type.includes("json") || format.includes("xml") && dataType.dataset.type.includes("xml")) {
+        dataType.hidden = false
```
halvorsanden, line 110: See comments above

js/api.js
Igor-J86: Initially I added them, and removed them.. But agree, it's better if they are there..! 😁

js/api.js
Igor-J86: Ah, right! Good catch! It seems to still be working fine then..! 😃

halvorsanden:
I see that the few examples in pickup point that has two instead of three formats are not showing any examples if javascript is selected. I think that’s OK, Checkout must add something if the have to. But json or xml is shown in the select. Should we clear the select in those instances where there is no match? (if that’s even possible)

Igor-J86:
> I see that the few examples in pickup point that has two instead of three formats are not showing any examples if javascript is selected. I think that’s OK, Checkout must add something if the have to. But json or xml is shown in the select. Should we clear the select in those instances where there is no match? (if that’s even possible)

How about setting a disabled attribute on the select then..?

halvorsanden:
> 
> How about setting a disabled attribute on the select then..?

Wouldn’t that make it not possible to use that to switch back to a different type?

js/api.js
Igor-J86: Refactored a bit now.. 🙂

halvorsanden:
> > > How about setting a disabled attribute on the select then..?
> > 
> > 
> > Wouldn’t that make it not possible to use that to switch back to a different type?
> 
> Oh, right... Doh... Hmm..

I think maybe something like `selectElement.selectedIndex = -1` works to unset the select.
https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedIndex

Igor-J86:
> > > > How about setting a disabled attribute on the select then..?
> > > 
> > > 
> > > Wouldn’t that make it not possible to use that to switch back to a different type?
> > 
> > 
> > Oh, right... Doh... Hmm..
> 
> I think maybe something like `selectElement.selectedIndex = -1` works to unset the select. https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedIndex

Thanks! That seems to do the trick! I think it works better now..! 😃

PR #1283, opened 2022-09-21
Use vite

arnemart:
Replace webpack with vite


js/main.js
```diff
@@ -0,0 +1,20 @@
+import "../css/main.css"
```
arnemart, line 1: Import the main css file here to make sure it is included in the build

js/main.js
```diff
@@ -0,0 +1,20 @@
...
+
+import AnchorJS from "anchor-js"
+const anchors = new AnchorJS();
+anchors.options = { placement: "left", visible: "hover" }
+anchors.add(".dev-anchored")
```
arnemart, line 6: Import Anchorjs as a dependency instead as a bundled file

js/main.js
```diff
@@ -0,0 +1,20 @@
...
+
+import "./servicefiltering.js"
+import "./api.js"
+import "./search/algolia.jsx"
+import "./scrollmarker.js"
```
arnemart, line 11: Load all js on all pages, the files aren't that huge (compared to the algolia one anyway, which is always loaded)

js/main.js
```diff
@@ -0,0 +1,20 @@
...
+
+import "./drawer.js"
+
+import "./tabs.js"
+window.tabs.init()
+
+import "./collapsible.js"
+window.collapsible.init()
```
arnemart, line 19: Include these in the build instead of loading them directly

js/search/AlgoSearch.jsx
```diff
@@ -9,7 +9,7 @@ const searchClient = algoliasearch(
 )
 
 const AlgoSearch = () => (
-  <InstantSearch searchClient={searchClient} indexName={process.env.ALGOLIA_INDEX || "devsite_qa"}>
+  <InstantSearch searchClient={searchClient} indexName={import.meta.env.ALGOLIA_INDEX || "devsite_qa"}>
```
arnemart, line 12: Vite includes environment variables in a different place

layouts/_default/baseof.html
```diff
@@ -5,13 +5,14 @@
     <body class="dev-docslayout">
       {{ partial "header.html" . }}
       {{ block "main" . }} {{ end }}
+      {{- partial "scripts.html" . -}}
```
arnemart, line 8: Scripts need to be loaded inside `<body>`

layouts/partials/scripts.html
```diff
@@ -1,7 +1,12 @@
-<script async src="{{ "/js/anchor.min.js?" | relURL }}{{ now.Unix }}" onload="window.anchors.options = { placement: 'left', visible: 'hover' }; window.anchors.add('.dev-anchored')"></script>
-<script async src="{{ "/js/tabs.js?" | relURL }}{{ now.Unix }}" onload="window.tabs.init()"></script>
-<script async src="{{ "/js/collapsible.js?" | relURL }}{{ now.Unix }}" onload="window.collapsible.init()"></script>
-<script async src="{{ "/js/drawer.js?" | relURL }}{{ now.Unix }}"></script>
-<script async src="{{ "/js/algolia.js?" | relURL }}{{ now.Unix }}"></script>
+{{ range sort $.Site.Data.manifest "file" "desc" }}
```
arnemart, line 1: The keys in the manifest.json file look like this: "js/main.js" and I could not figure out how to reference keys like that without iterating over the whole map, it's not very nice but it works. The sorting is to make sure the polyfill chunk is loaded before the main chunk on legacy browsers.

vite.config.js
```diff
@@ -0,0 +1,25 @@
...
+  },
+  plugins: [
+    legacy(), 
+    react(),
+    viteStaticCopy({
+      targets: [{
+        src: 'static/assets/manifest.json',
+        dest: '../../data'
+      }]
+    })
```
arnemart, line 23: Copy the manifest file into hugos `data` directory after each build. This is only necessary to make live reloading work when running `npm start` and `npm bundlewatch` together, if the manifest file lives outside the data directory hugo will not pick up changes in it and will load try to load the wrong js files when vite rebuilds.

layouts/partials/scripts.html
```diff
@@ -1,7 +1,18 @@
-<script async src="{{ "/js/anchor.min.js?" | relURL }}{{ now.Unix }}" onload="window.anchors.options = { placement: 'left', visible: 'hover' }; window.anchors.add('.dev-anchored')"></script>
-<script async src="{{ "/js/tabs.js?" | relURL }}{{ now.Unix }}" onload="window.tabs.init()"></script>
-<script async src="{{ "/js/collapsible.js?" | relURL }}{{ now.Unix }}" onload="window.collapsible.init()"></script>
-<script async src="{{ "/js/drawer.js?" | relURL }}{{ now.Unix }}"></script>
-<script async src="{{ "/js/algolia.js?" | relURL }}{{ now.Unix }}"></script>
+{{ if $.Site.IsServer }}
+  <script type="module">
```
arnemart, line 2: This is the preamble code for react refresh. Vite would normally stuff this into the html, but since we're not loading html with vite we need to include it ourselves otherwise vite would complain.

package.json
```diff
@@ -4,16 +4,18 @@
...
+    "bundlewatch": "vite build --watch",
     "buildraml": "node scripts/raml2json",
     "buildexamples": "node scripts/buildApiExamples",
     "build": "npm run bundle && npm run buildraml && npm run buildexamples",
     "buildqa": "npm run clean && npm run build && HUGO_ENV=qa hugo",
     "justhugo": "HUGO_ENV=qa hugo",
     "buildprod": "npm run clean && npm run build && HUGO_ENV=production hugo",
-    "start": "cross-env HUGO_ENV=dev hugo server --ignoreCache",
-    "bundlewatch": "webpack watch --mode development",
+    "start": "concurrently --kill-others npm:start-*",
```
arnemart, line 16: Run the vite dev server and the hugo dev server concurrently! How nice

halvorsanden (approved):
Very clever to comment on the special parts - made the changes easier to understand.

js/search/AlgoSearch.jsx
arnemart: UPDATE: this does not work, see https://github.com/bring/developer-site/pull/1294

PR #1292, opened 2022-09-23
Remove post-css-variables config and CSS custom props

Igor-J86:
Removing post-css-variables config and several CSS custom properties (CSS variables) as custom properties are now globally available from the [Mybring Design System](https://github.com/bring/mybring-design-system).

You will find all of them listed here:
[Mybring Design System - Custom Properties](https://www.mybring.com/design-system/custom-properties/)


bringdev:
[trello.com/c/lhYXZqCB](https://trello.com/c/lhYXZqCB)

arnemart (approved):
Looks good!

PR #1294, opened 2022-09-27
Algolia cleanup

arnemart:
Make sure algolia uses the correct index, and also that empty searches don't use up our api quota

### DIFF
--- a/js/search/AlgoSearch.jsx
+++ b/js/search/AlgoSearch.jsx
@@ -1,12 +1,33 @@
-import React from 'react'
-import algoliasearch from 'algoliasearch/lite'
-import { InstantSearch } from 'react-instantsearch-dom'
-import Result from './Result.jsx'
+import React from "react"
+import algoliasearch from "algoliasearch/lite"
+import { InstantSearch } from "react-instantsearch-dom"
+import Result from "./Result.jsx"
 
-const searchClient = algoliasearch(
-  'PWK5X2R2GD',
-  '75cb7755dd0c05b3f3629e7a6793a2a3'
+const algoliaClient = algoliasearch(
+  "PWK5X2R2GD",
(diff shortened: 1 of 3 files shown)


js/search/AlgoSearch.jsx
```diff
@@ -1,15 +1,36 @@
...
+    if (requests.every(({ params }) => !params.query)) {
+      return Promise.resolve({
+        results: requests.map(emptyAlgoliaResponse)
+      })
+    }
+
+    return algoliaClient.search(requests)
+  }
+}
+
```
arnemart, line 31: This is very convoluted but it's the official way of detecting and ignoring empty search queries.

vite.config.js
```diff
@@ -3,6 +3,9 @@ import legacy from "@vitejs/plugin-legacy"
 import react from "@vitejs/plugin-react"
 
 export default defineConfig({
+  define: {
+    __ALGOLIA_INDEX__: `"${process.env.ALGOLIA_INDEX || "devsite_qa"}"`,
```
arnemart, line 7: This is "global constant replacement", and will replace every instance of __AGOLIA_INDEX__ in the built source code with the correct quoted algolia index name.

js/search/AlgoSearch.jsx
```diff
@@ -1,15 +1,36 @@
...
 const AlgoSearch = () => (
-  <InstantSearch searchClient={searchClient} indexName={import.meta.env.ALGOLIA_INDEX || "devsite_qa"}>
```
arnemart, line 12: For future reference, why this did not work: https://vitejs.dev/guide/env-and-mode.html
> To prevent accidentally leaking env variables to the client, only variables prefixed with VITE_ are exposed to your Vite-processed code

Igor-J86 (approved):
LGTM! 😃 🔍

PR #1309, opened 2022-10-03
Fix conflicting data-attributes and IDs on filters

Igor-J86:
At first glance the filtering on Service Portfolio page seemed buggy, but after a small investigation it became clear that the reason for this behaviour was "conflicting" data-attributes and IDs between the "Booking and Shipping Guide (BSG)" and "Value added Services (VAS)".

A small adjustment on the VAS data-attributes and IDs is made with this PR. 🙂

### DIFF
--- a/layouts/partials/services/vas.html
+++ b/layouts/partials/services/vas.html
@@ -10,9 +10,9 @@
         <div class="mrl">
           <p class="form__label">Filter by</p>
           <div class="btngroup">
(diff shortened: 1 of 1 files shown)


bringdev:
[trello.com/c/TkuX2EAW](https://trello.com/c/TkuX2EAW)

halvorsanden (approved):

PR #1310, opened 2022-10-03
Show only first request example

Igor-J86:
If there are multiple request examples, they were all displayed on page load. With this PR only the first will be displayed (as it is for the response examples).


bringdev:
[trello.com/c/aJ3SahKu](https://trello.com/c/aJ3SahKu)

halvorsanden (approved):

PR #1311, opened 2022-10-04
Split service portfolio into two files

halvorsanden:
Since it’s too difficult to keep both sheets in one file …
The entire xlsx was overwritten by the last update to Booking and Shipping Guide, instead of updating only the relevant sheet – so we’ll just avoid having them in the same file as reports and invoice.


bringdev:
[trello.com/c/TExWS6PF](https://trello.com/c/TExWS6PF)

Igor-J86 (approved):
LGTM 😃

PR #1315, opened 2022-10-04
Add aria attributes to the response schema param buttons

Igor-J86:
Adding the `aria-expanded` and `aria-controls` attributes on the response schema param buttons, signifying the expanded (or not expanded) state of the button, and what element the button controls.
The `aria-controls` attribute targets the matching ID of the element that the button controls, so have to move the ID to the element itself.

The JS had to also be modified a bit to accommodate `aria-expanded` state on the buttons.

### DIFF
--- a/js/api.js
+++ b/js/api.js
@@ -3,11 +3,13 @@ const paramToggleBtn = document.querySelectorAll(".param-toggle-btn")
 
 paramToggleBtn.forEach((btn) => {
   btn.addEventListener("click", function () {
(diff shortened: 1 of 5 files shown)


bringdev:
[trello.com/c/uIJ7dZg7](https://trello.com/c/uIJ7dZg7)

halvorsanden (approved):
Very nice use of aria in the JS and switch of the IDs 👍

PR #1323, opened 2022-10-07
Improve header

Igor-J86:
At first glance it can look like not much has changed, but here are the details:

- "Blog" has been removed from the header navigation. 
- The dark mode / light mode switch has been refactored into a button with a custom dropdown, containing the existing buttons for switching between auto / light / dark. This is making the switch less prominent and saving some space. It has also been moved to after the search field.
- The search icon has been moved "inside" the input field, which is additionally saving some space.
- With these adjustments it made way to add some breakpoints in CSS for the different parts of the header, making them stack better on smaller screens. It's maybe not perfect, but it looks much better this way in my opinion.


bringdev:
[trello.com/c/I3YqRhIq](https://trello.com/c/I3YqRhIq)

halvorsanden:
Can we move the search field closer to the navigation items?
And now that the theme options are in a dropdown, should we rename the button to "theme"? And have the same hover effect as the one one the logo? And also remove text-note form the options?

I guess that’s a lot of detail management posed as questions … feel free to disagree though.

Igor-J86:
> 

- I tried placing the search field closer first, but it only resulted in a strange-looking header on wider screens. But I will move it a bit further to the right, along with the theme switcher.
- Agree on adding the same hover on the theme switcher, as it is on the logo. Although, renaming the button only to "theme", I feel that it will loose the "explanation" of what you have selected. In this way it's quite clear what you have chosen (if you manually choose something). 
- Oh, yes! Removing the text-note on the options as well..

halvorsanden:
>
>  I tried placing the search field closer first, but it only resulted in a strange-looking header on wider screens. But I will move it a bit further to the right, along with the theme switcher.
> 

Not sure if I see any layout strangeness in removing `margin-right: auto` from the search field. I think it tidies it up a bit versus having it in the centre.

Igor-J86:
> > I tried placing the search field closer first, but it only resulted in a strange-looking header on wider screens. But I will move it a bit further to the right, along with the theme switcher.
> 
> Not sure if I see any layout strangeness in removing `margin-right: auto` from the search field. I think it tidies it up a bit versus having it in the centre.

Ok, we can go with that.. It was something that didn't feel right, but looking at it again now I agree it looks a bit tidier. Will do the adjustment. 😃

halvorsanden (approved):

PR #1333, opened 2022-10-12
Check whether slack webhook allows channels other than #deploy

arnemart:


arnemart:
Nope, that does not work. Good to know.

PR #1336, opened 2022-10-12
Allow booleans as default param values

halvorsanden:
This checks for the existence of the value and displays it – even if the value itself is `false`
Ref: https://pkg.go.dev/fmt


Igor-J86 (approved):
True that something was...false... 🤭

PR #1337, opened 2022-10-13
Add light mode css for examples

Igor-J86:
Adding light mode CSS in the syntax.css, following the same logic as we have in place for the dark theme. So, when a user have dark mode enabled or selected, the original dark mode syntax highlighting CSS is in use, otherwise the newly added light mode CSS will be in use.

We can generate a syntax highlighting CSS with Hugo. In order to do so, we need to install Hugo locally.

Quick start:
https://gohugo.io/getting-started/quick-start/

Then we can generate one base on the available styles:

`hugo gen chromastyles --style=monokai > syntax.css`

Gallery of available styles.
https://xyproto.github.io/splash/docs/longer/all.html

I chose **rainbow_dash**, but modified it so the colours don't pop too much and have some balance between dark mode and light mode.

Chosen theme:
https://xyproto.github.io/splash/docs/longer/rainbow_dash.html

JSON example:
Skjermbilde 2022-10-13 kl  10 30 04

XML example:
Skjermbilde 2022-10-13 kl  10 33 31

### DIFF
--- a/css/base.css
+++ b/css/base.css
@@ -82,7 +82,7 @@ samp {
 
 pre {
   background-color: var(--gray-95);
-  font-family: Monaco, Consolas, "Lucida Console", monospace;
+  font-family: Consolas, SF Mono, Menlo, 'Andale Mono', 'Ubuntu Mono', 'Courier New', Courier, monospace;
   font-size: 0.82rem;
   padding: 0.75rem;
   overflow: auto;
--- a/css/code.css
+++ b/css/code.css
@@ -11,8 +11,6 @@
 code,
 .filename {
   white-space: pre;
-  font-family: Consolas, Monaco, "Andale Mono", "Ubuntu Mono", "Courier New",
-    Courier, monospace;
   font-size: 0.82rem;
 }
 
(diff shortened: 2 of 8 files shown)


bringdev:
[trello.com/c/wLhLDE4G](https://trello.com/c/wLhLDE4G)

halvorsanden:
I have mixed success with bold code fonts – but maybe it works better if we update the code font stack in MDS?
Screenshot 2022-10-13 at 10 39 31

Igor-J86:
> I have mixed success with bold code fonts – but maybe it works better if we update the code font stack in MDS? Screenshot 2022-10-13 at 10 39 31

Oh, right.. This is the `font-weight: 700;` causing it, it seems...? I think maybe updating the code font stack in MDS will solve this, but we could also remove the `font-weight: 700;` from the syntax highlighting CSS as well. I don't see whether it is beneficial to have it there or not though...

halvorsanden (approved):
Feel free to merge, whether you remove the font-weight or look at the font stack first.
… and good work! 👍😁

PR #1346, opened 2022-10-25
Warehousing-related template adjustments

halvorsanden:
- Align request type at the top of the url box
- Not show endpoint section if there are none
- Some Warehousing related fixes
  - Move intro content into intro section
  - Use mb-table class
  - Remove custom "required" class, use mb-badges instead
  - Fix dead link and unncessary use of table tags
 
No content has been changed, just moved around.

### DIFF
--- a/content/api/warehousing3/index.html
+++ b/content/api/warehousing3/index.html
@@ -28,20 +28,17 @@ information:
 documentation:
   - title: Order Information
     content: |
-      In order to obtain information about an order/get the receipt for an order, you have two choices:
+      In order to obtain information about an order/get the receipt for an order, you have two options:
       <head>
         <link rel="stylesheet" href="warehousing.css">
         <script src="listcollapse.js" type="text/javascript" language="javascript1.2"></script>
         <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.3.1/jquery.min.js"></script>
       </head>
       <body>
-        <p></p>
         <ul>
-          <li>You can obtain data from Mybring.</li>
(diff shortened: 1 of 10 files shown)


bringdev:

content/api/warehousing3/index.html
```diff
@@ -20,29 +26,19 @@
...
-          <li>You can obtain data from Mybring.</li>
-          <table class="codeBox"><tr><td class="dark">GET</td><td><a href="https://developer.bring.com/api/warehousing/#list-order-information">https://developer.bring.com/api/warehousing/#list-order-information</a></td></tr></table>
-          </ul>
+          <li>You can <a href='{{< relref "warehousing4" >}}#endpoints'>obtain data from Mybring</a>.</li>
           <li>We can push data to you if you set up a webservice on your side</li>
-        </ul><br>
+        </ul>
         <p>You should use a REST web service which needs to be secured through https.</p>
         <p>After you have set up your web service, you must send the information about the URL and its authentication to us, so that we can set up a link from our side.</p>
         <p>The data sent from Bring will be in GS1 3.3 format.
```
Igor-J86, line 44: Should the "Authentication" and "Verification" (below here) also be move up to the details/summary elements?

halvorsanden: Yes, ideally, they should be in the same place but I struggle to understand what it’s saying, so I didn’t want to move it in case it was not related to that other text or needed more editing. I think we should have content improvements as a point in our card, but that must involve the team that owns the API – so it is correct and they know what’s going on.

Igor-J86: Agree 🙂

Igor-J86 (approved):
This already shapes things up a bit! 😃

PR #1347, opened 2022-10-26
Don't put a dash after the prefix if there is no prefix

arnemart:
Not sure if this is the best way of fixing this but it seems to work?


bringdev:
[trello.com/c/whGhTYul](https://trello.com/c/whGhTYul)

Igor-J86 (approved):
Seems to be working fine! 👍🏼

PR #1360, opened 2022-11-08
Bring environment info webcomponent

Igor-J86:
A webcomponent for Bring environmental information that can be served from our dev-site. The script can be linked to externally, and by only using the custom element `<bring-environmental-info></bring-environmental-info>`, the information from the script will be populated in that custom element with its own shadow dom, making the component encapsulated so external CSS or JS don't get in conflict and affect the content provided by us.

**As modal window (optional):**
By adding `modal="true"` on the custom element, this information will appear in a modal window instead. The modal window can be opened by adding the data attribute `data-bring-env-open` on an element.

Vite config also needs an adjustment so the filename don't get a hash-code in the filename. We could use the [build.lib](https://vitejs.dev/config/build-options.html#build-lib) option, but this is not supported when using the legacy plugin.

**Further improvements**
- Possibility to show in different languages, by optionally passing a language code by an attribute (?)
- Setting a Posten or Bring theme by an attribute (?)


bringdev:
[trello.com/c/e2zFW5uB](https://trello.com/c/e2zFW5uB)

content/api/e-commerce-solutions/environmental-info/index.html
```diff
@@ -0,0 +1,48 @@
...
+    title: Environmental info
+weight: 2
+---
+
+<section class="dev-docscontent__section">
+  <p class="text-heading">
+    A web component where you can easily present the environmental information
+    from Posten/Bring in your e-commerce. The information can be displayed by adding
+    the script served by us and the custom html element provided below. Optionally a modal
+    window can be triggered where the information will be displayed instead.
```
halvorsanden, line 18: Putting the value even more upfront. Then explanation and options, and then leading into _how_.

```suggestion
    Present carbon offset information from Posten and Bring in your e-commerce solution. 
    A custom web component lets you present the information embedded on a page or in a 
    modal window. This is all done through a script served by us and the custom HTML 
    element as described below.
```

js/webcomponent/bring-env/style.js
```diff
@@ -0,0 +1,212 @@
...
+      ${isModal ?
+      `height: calc(100% - 5rem);
+      overflow-y: auto;`
+      : ''}
+    }
+
+    .bring-env section {
+      display: flex;
+      flex-direction: column;
+      width:100%;
```
halvorsanden, line 137: Section is a block element, so it should be 100 % wide in modern browsers. Meaning that the flex also isn’t necessary to stack it in a column. …unless there’s some responsiveness coming into play?

js/webcomponent/bring-env/style.js
```diff
@@ -0,0 +1,212 @@
...
+    : ''}
+
+    ${isModal ? `
+      .bring-env section:last-child {
+        margin-bottom: 4rem;
+      }`
+    : ''}
+
+    .bring-env-text {
+      width: 100%;
```
halvorsanden, line 158: ref. comment above. If they are not made into flex children, I don’t think they need this setting.

js/webcomponent/bring-env/style.js
```diff
@@ -0,0 +1,212 @@
...
+      margin-top: 1rem;
+    }
+
+    .bring-env-cards .bring-env-card {
+      background: #fff5f0;
+      border-radius: 3px;
+      padding: 2rem;
+      flex: 1 1 15rem;
+      display:flex;
+      flex-direction: column;
```
halvorsanden, line 181: Seems to also work without these two lines since p are block elements.

js/webcomponent/bring-env/style.js
```diff
@@ -0,0 +1,212 @@
...
+      text-align: center;
+    }
+
+    .bring-env-cards .bring-env-card svg {
+      max-height: 3rem;
+      display: block;
```
halvorsanden, line 187: Or you can keep the flex above and use `flex: 0 0 3rem;` instead on this.

halvorsanden (commented):
A couple of flex things that can be cut – we can look at those on Monday, not sure what’s best to do. Just a few lines really, but probably nice to keep it as simple as possible.

js/webcomponent/bring-env/style.js
Igor-J86: True! Will adjust the CSS a bit with regards to this..

content/api/e-commerce-solutions/environmental-info/index.html
Igor-J86: Oh, great! Much better! 🙌🏻

js/webcomponent/bring-env/style.js
Igor-J86: Removing this as well..

js/webcomponent/bring-env/style.js
Igor-J86: The `flex: 0 0 3rem;` is ment for the width, but on the SVGs I wanted to make sure they stay the same height, so therefore the `max-height: 3rem;`.

halvorsanden (approved):
🤩
I agree on the language and themes – cool possibilities enables by the web component.

PR #1365, opened 2022-11-11
subpage experiment

tomgag:
- Add static content from existing site
- WIP


halvorsanden:
Move `new.md` into the `new` folder and rename the file to `_index.md`
That will give the same structure and urls, but contain NEW in a separate folder. The low dash in the filename allows subpages to be made.

content/api/booking/new/rfid.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Mailbox Parcel, 3570 (tracking with RFID)
+layout: api
+hidden: true
+notanapi: true
+
+documentation:
```
halvorsanden, line 7: I think this should be:

```suggestion
information:
```

But I don’t get it to work – I’ll take a look at the template …

halvorsanden: No, wait, that’s not correct … you were right 🙈

PR #1370, opened 2022-11-15
E-commerce solutions adjustments

Igor-J86:
Changing the sort order of the e-commerce solutions, where the "Checkout guide Norway" and "Checkout request builder" comes first, before the "Product page widget" and "Environmental info". 
"Checkout guide Norway" is also then replacing the "Product page widget" on the front page, and adding a card for the newly created "Environmental info" web component on the E-commerce solutions page.

Additionally, the icon for Order Management is replaced by the same icon that is in use for OM inside Mybring.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -59,15 +59,20 @@ tag = "tags"
 [menu]
 
 [[menu.api]]
-title = "API operational status"
-url = "https://status.bring.systems"
-weight = 100
-
-[[menu.api]]
+identifier = "checkout-request-builder"
 title = "Checkout request builder"
 url = "https://www.mybring.com/shipping-guide/integration/checkout"
 parent = "e-commerce-solutions"
-weight = 100
+weight = 2
+
+[menu.api.params]
+blank = "target=\"_blank\" rel=\"noopener noreferrer\""
+
+[[menu.api]]
+identifier = "api-operational-status"
+title = "API operational status"
+url = "https://status.bring.systems"
+weight = 7
 
 [menu.api.params]
 blank = "target=\"_blank\" rel=\"noopener noreferrer\""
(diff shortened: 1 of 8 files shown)


config.toml
```diff
@@ -59,15 +59,20 @@ tag = "tags"
...
-[[menu.api]]
+identifier = "checkout-request-builder"
 title = "Checkout request builder"
 url = "https://www.mybring.com/shipping-guide/integration/checkout"
 parent = "e-commerce-solutions"
-weight = 100
+weight = 2
+
+[menu.api.params]
+blank = "target=\"_blank\" rel=\"noopener noreferrer\""
```
halvorsanden, line 69: Duplicate of L77-78?

halvorsanden (approved):
Just one comment. 😊

config.toml
Igor-J86: I tried putting it only once, but then it only works for the menu item that is mentioned above it. When I put it after each menu item, then it's also picked up in the template for each menu item. Is it maybe another way to do it..?

halvorsanden: I thought it should work more or less as a variable, but it’s a while since it was added so I have to look into it …

halvorsanden: Just me who can’t read TOML. The param belongs to the individual entry.

As far as I can see, it equals:

```
[[menu.api]]
identifier = "checkout-request-builder"
title = "Checkout request builder"
url = "https://www.mybring.com/shipping-guide/integration/checkout"
parent = "e-commerce-solutions"
weight = 2
blank = "target=\"_blank\" rel=\"noopener noreferrer\""
```

I say we convert the config to JSON. I’d say we are more familiar with reading that. But feel free to merge this 👍

Igor-J86: Oh right, ok.. Thanks! 👍🏼

PR #1371, opened 2022-11-15
Convert config from toml to json

halvorsanden:
- Makes the config easier to read for us and the scripts
- We can remove a dependency
- I have compared the output from the two scripts before and after; they are identical
- The big diff in the package-lock is because it had been converted to an lower version by someone running an old version of Node. (The _someone_ has been notified 🤓)


bringdev:
[trello.com/c/g8R3VvGl](https://trello.com/c/g8R3VvGl)

Igor-J86 (approved):
I like this! Much neater and easier to read..! 😃

PR #1373, opened 2022-11-17
Enable use of $ref in request examples

halvorsanden:
It’s more or less the same as the response file, minus 20 lines of code because there is no need for different response codes in the request.

This is needed for the upcoming Shipping Guide OAS version.


bringdev:
[trello.com/c/mo69aCdO](https://trello.com/c/mo69aCdO)

Igor-J86 (approved):
🚀

PR #1374, opened 2022-11-18
Support request body schema

halvorsanden:
More or less the same templates as for the response, but without the response code levels and the background. They could possibly be using the same template with some reworking, so I’m adding that to the card for cleanup that we have.

Turns out that Modify Shipment API has had this all along, so that gives a small example. Otherwise, I’ve tested the switching between media types and it looks like this for Shipping Guide:
Screenshot 2022-11-17 at 21 04 37
Screenshot 2022-11-17 at 21 04 54

### DIFF
--- a/layouts/_default/api.html
+++ b/layouts/_default/api.html
@@ -72,8 +72,59 @@
                     </div>
                     <p class="mbl text-heading">{{ .description | safeHTML }}</p>
 
+                    {{- with or .parameters .requestBody -}}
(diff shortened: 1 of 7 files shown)


bringdev:
[trello.com/c/mo69aCdO](https://trello.com/c/mo69aCdO)

Igor-J86 (approved):
Looks good! 😃

PR #1377, opened 2022-11-30
Fix xml example problem for shipping guide docs

Igor-J86:
Setting `string` on `$value` assumes that the `$value` is a single value, like a string or integer. After some investigations I found out that the `$value` was suddenly an array in a case, so to fix the issue we can check if `$value` is an array, loop over it, and output each value from the array as a xml line. Otherwise, just output the xml line as before.


bringdev:
[trello.com/c/HhtGXCJB](https://trello.com/c/HhtGXCJB)

halvorsanden (approved):

PR #1384, opened 2022-12-02
Enable api docs subpages with subnavigation

halvorsanden:
This creates a subnav in the sidemenu, a backward nav on the subpages and a subpage list on the main API page. This is done by adding frontmatter as described in the readme.

Not completely satisfied with the duplication of the menu code, but it’s what we have time for now – will add a cleanup to a card.

The "bookings" API is just a test. We don't have any data on this in master yet, only a screenshot.

Screenshot 2022-12-02 at 12 50 52

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,30 @@ oas: {Link to OAS JSON, typically outputted by Swagger}
 ---
 ```
 
+## Adding subpages to an API
+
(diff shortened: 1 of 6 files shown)


bringdev:
[trello.com/c/UnfGB8Za](https://trello.com/c/UnfGB8Za)

Igor-J86 (approved):
LGTM 😁

PR #1393, opened 2022-12-14
Consolidate request example templates

halvorsanden:
- Checks each request example if it’s a component reference or not, instead of doing a general check for the existence of components and assuming that all examples are using `$ref`
- Adds a heading if there’s only one example (except if that heading is just "example")
- Reduces the complexity and number of template files (will look into doing the same for the response examples and possibly joining them all into one – but I doubt that is efficient)

Example in api/bookingnew

Ignore the master merge commits – I got some local git setup issues to fix


bringdev:

Igor-J86 (approved):
A really nice improvement! 😃

PR #1394, opened 2022-12-16
Consolidate and bugfix response examples

halvorsanden:
- Fixed a bug introduced by https://github.com/bring/developer-site/pull/1393 where the schema was shown in the response examples.
- Checks each request example to see if it’s a component reference or not instead of doing a general check for the existence of components and assuming that all examples are using $ref
- Adds a heading if there’s only one example (except if that heading is just "example" or "response")
- Reduces the complexity and number of template files (I will return to the request example and improve that further)

Known issue: when switching to XML, all non-XML examples disappear – meaning that some places are left blank.


bringdev:

PR #1396, opened 2022-12-19
Example tuning

halvorsanden:
- Make singular example title background dark like when there are multiple
- Add spacing between example types
- Make soap examples response code display similar to the rest
- Make sidemenu top padding match content, so they start on the same place.

Screenshot 2022-12-19 at 13 31 22

### DIFF
--- a/css/documentation.css
+++ b/css/documentation.css
@@ -153,6 +153,10 @@
   width: 38%;
 }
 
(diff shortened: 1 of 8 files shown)


bringdev:

Igor-J86 (approved):

PR #1397, opened 2022-12-20
Use one template for OAS endpoints

halvorsanden:
Much of the code in restsoap.html was duplicated from api.html for rendering OAS endpoints. It was outdated (already) so this extracts the newest code to a separate endpoint.html partial for reusability.

The check for APIs with rest and soap was based on the page title – it now checks the page’s unique identifier, similar to what’s done in the menu. It’s also placed inside the oas check, so it’s possible to have rest-soap WIP as well.


bringdev:
[trello.com/c/xT8vFTba](https://trello.com/c/xT8vFTba)

cdwinsnes (approved):
nice to get rid of Title as identifier 👍🏻

PR #1400, opened 2022-12-29
Improve accessibility on Environmental info modal

Igor-J86:
Improving the accessibility on Environmental info modal by adding `aria-modal="true"` in conjunction with `role="dialog"` (if it's a modal). The focus is also moved from the trigger when opening the modal, to the close button in the modal, so the user can continue in the modal, or easily close it. 
In addition, support is added for pressing the ESC key and closing the modal.
When the modal is closed, the focus is moved back to the trigger.

### DIFF
--- a/js/webcomponent/bring-env/main.js
+++ b/js/webcomponent/bring-env/main.js
@@ -45,11 +45,12 @@ class BringEnv extends HTMLElement {
         closeModal.addEventListener('click', () => {
           close();
         });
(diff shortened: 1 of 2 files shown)


bringdev:
[trello.com/c/1aBNvAeZ](https://trello.com/c/1aBNvAeZ)

halvorsanden (approved):

PR #1405, opened 2023-01-03
Example stability fixes

halvorsanden:
- Prevents rendering of example response buttons when there is no corresponding example (it only did a general example check before assuming that all responses had examples if only one of them had one)
- Makes the same button’s id unique (it was only used by the corresponding tabpanel for aria-label)
- Enables reuse of examples within a response code by adding an index to the rendered example id


bringdev:

Igor-J86 (approved):
LGTM 😃

PR #1408, opened 2023-01-04
Colour and aria fixes

halvorsanden:
- Fixes 20 broken aria references in tracking where the aria-labelledby on the tabpanel didn’t match the id of the tab in the oneOf schema.
- Removes the tab roles from the oneOf response schema (doing the request schema separately as part of the shipment API improvement) and just connect the radio input to the description list with aria-controls and describing the list with aria-label.
- Improves colour contrast on the examples buttons plus a couple of other small things.

### DIFF
--- a/layouts/partials/api/oas/response-schema-xml.html
+++ b/layouts/partials/api/oas/response-schema-xml.html
@@ -44,13 +44,13 @@
   {{- if eq $isOneOf "radio" -}}
     {{- $levelId = printf "%s-%s" $levelName "oneof-xml" -}}
     <div>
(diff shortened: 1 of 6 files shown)


bringdev:

layouts/partials/api/oas/response-schema-xml.html
```diff
@@ -42,13 +42,15 @@
   
   {{/*  Render oneOf radio inputs  */}}
   {{- if eq $isOneOf "radio" -}}
+    {{- $levelId = printf "%s-%s" $levelName "ofone-xml" -}}
```
Igor-J86, line 45: Should "ofone" maybe be named to "oneOf" as it is in the rest of the code..?

halvorsanden: Turned out that role=tab on radio input isn’t really necessary – don’t know why I started overwriting the default semantics in the first place 🤷‍♂️

Igor-J86 (approved):

PR #1411, opened 2023-01-05
Algolia label

halvorsanden:
The implementation uses [react-autosuggest](https://github.com/moroshko/react-autosuggest) that doesn’t come with the basic functionality of a label 🤯

We can look into throwing it out, possibly building our own implementation for search and autosuggest, but for now I have added a hidden label.


bringdev:

Igor-J86 (approved):
Ah, of course! Nice! 😃
Seems like I over-complicated stuff when looking into it.. 😵‍💫

halvorsanden:
> Ah, of course! Nice! 😃 Seems like I over-complicated stuff when looking into it.. 😵‍💫

Blame the component package, pre-made stuff has a tendency to lead to creative workarounds. 🙃

PR #1413, opened 2023-01-05
Fix heading levels

halvorsanden:
Just a handful.
There were a couple more that I didn’t fix, they will be gone as soon as the old documentation is replaced.


bringdev:

Igor-J86 (approved):

PR #1423, opened 2023-01-11
Render req example value correctly with 3 strings

halvorsanden:
Leaving out a string renders the string but with some extra text. Either way the value of the select option is wrong and didn’t match the element it was meant to control.

This only affected request examples that were not using $ref.


pedromacli (approved):
LGTM 👍

PR #1428, opened 2023-01-13
Add expand and collapse all buttons for schemas

halvorsanden:
Small things that turn out to not be that small …

Adds two buttons because neither really reflect state. For example, it’s possible to expand all and then collapse individual ones.
I tried to make keep it simple – there’s always the possibility to check every media type for levels and just not render the buttons at all, but I ended up hiding them as default and use JS to show them based for schemas that have a total of more than one toggle button.
Most of the time, if they have one toggle, it’s the first XML level which is open by default – or if there’s only JSON, one single toggle doesn’t need three buttons to do the same.

Responsiveness was probably the most tricky to solve.

The request and response schemas had slightly different text sizes and the "required" tag was styled differently. So that’s fixed now. Room for improvement on that still – some of the text looks a bit too small, but let’s return to it if it bugs us over time.

The toggle button had a bug, but it wasn’t visible until looking at some of the new, unpublished docs, where one toggle remained had mismatch between its state and the aria-expanded, rendering the toggle unresponsive as the JS operate with the aria. Fixing it while reducing the two template files into one.

Screenshot 2023-01-13 at 15 37 55

### DIFF
--- a/css/documentation.css
+++ b/css/documentation.css
@@ -259,3 +259,8 @@ div.response {
   margin-bottom: 0;
   flex: 1 1 8rem;
 }
+
+.schema__header:has(+ .schema__container .param-toggle-btn)
+  button[data-toggle-all] {
+  display: flex;
+}
--- a/css/documentation.css
+++ b/css/documentation.css
@@ -253,7 +253,7 @@ div.response {
   margin-top: 0;
 }
 
-.schematype-select {
+.schematype__select {
   padding: 0.25rem;
   height: auto;
   margin-bottom: 0;
--- a/css/documentation.css
+++ b/css/documentation.css
@@ -252,3 +252,10 @@ div.response {
 .example-wrap :first-child {
   margin-top: 0;
 }
+
+.schematype-select {
+  padding: 0.25rem;
+  height: auto;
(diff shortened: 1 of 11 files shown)


bringdev:
[trello.com/c/PD8A2LXp](https://trello.com/c/PD8A2LXp)

halvorsanden:
@Igor-J86 and I experimented with [has()](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) for showing the toggle all buttons. The solution works almost exactly like the JS version except:
- It only checks for the existence of toggle buttons, not quantity
- It doesn’t work in Firefox (but support for has() is coming within a few months by the look of things)

We decided to keep both and will remove the JS when CSS is supported better.

The CSS works like this:

```css
.schema__header:has(+ .schema__container .param-toggle-btn)
  button[data-toggle-all] {
  display: flex;
}
```

If `schema__header` has succeeding sibling (+) with class name `schema__container` which contains an element with class `param-toggle-btn`, then all elements with the selector `button[data-toggle-all]` inside `schema__header` will display as flex.

This is 16 lines fewer than the JS.

PR #1435, opened 2023-01-23
Granular API updates features

Igor-J86:
Splitting up the signup for API updates, so the visitor can choose to sign up either for all updates, or choose to sign up only for certain API areas. This means that the changelog entries (.md files) have to be structured in their own folders, that are named based on the different API areas (Booking, Checkout, Warehousing etc.).

If a changelog entry is added to the main folder, named `revision-history`, this is intended as a place where general updates can be placed. These updates will then be presented on the "general page" along with all other API updates from the different API areas, but also on each API area among the entries that are already there.

The RSS template follows the same logic, and we will use the different RSS urls to set up different segments with rules, so the signed up users will receive all updates or only from one API area, or several API areas, that they have selected and signed up for. These preferences can then be adjusted by the user later as well, either by selecting one of the other options on the dev-site and submitting their email again, or by clicking the "Edit preferences" link in the emails they receive.

**The signup form**
Skjermbilde 2023-01-20 kl  10 29 50

Skjermbilde 2023-01-20 kl  10 29 13

**The general revision history page**
Skjermbilde 2023-01-23 kl  15 50 35

**An API area, containing the updates from the general page**
Skjermbilde 2023-01-23 kl  15 50 57


bringdev:
[trello.com/c/lp2TOVPS](https://trello.com/c/lp2TOVPS)

halvorsanden (approved):

PR #1438, opened 2023-01-25
Convert Shipment API doc to OAS JSON

halvorsanden:
Replaces the RAML with a JSON source so that we can use the new templates. Should also work with a remote Swagger output as long as it’s made with OAS 3 spec.


bringdev:
[trello.com/c/682n6wbZ](https://trello.com/c/682n6wbZ)

PR #1455, opened 2023-02-02
Update developer site documentation for Invoices.

Oeitho:


kiangol:
Looks OK, but response examples are missing. I can only see the schema 🤔

halvorsanden:
Endpoint titles are also missing.
I can look into if we can avoid having separate entries for each media type extensions.

PR #1458, opened 2023-02-09
Fix template width for warehousing

halvorsanden:
- Add workaround for Warehousing, the email signup made the entire page narrower because everything was added as HTML inside the documentation section – which led to a some horizontal scrolling inside a table or something.
- Format some low hanging stuff like the unstyled buttons and the type which looked like links.
- Remove head tag from the middle of the page
- Use only one warehousing.css instead of three
- "fix" some indentation


bringdev:

Igor-J86 (approved):

PR #1460, opened 2023-02-10
Add API area to RSS feed titles

Igor-J86:
Adding the API area to RSS feed titles, if the param `apiArea` is present. This title is also read by Mailchimp for sending the campaigns that are set up based on the RSS feeds for the Bring API updates.

The general RSS (all API updates combined):
- Bring API updates

Each API area RSS:
- Bring API updates - `param.apiArea` APIs

**Examples:** 
Bring API updates - Checkout APIs
Bring API updates - Reports and invoices APIs

### DIFF
--- a/layouts/_default/rss.xml
+++ b/layouts/_default/rss.xml
@@ -31,10 +31,14 @@
 {{- $latestPage = . }}
 {{- end -}}
 {{- $pLink := .Permalink -}}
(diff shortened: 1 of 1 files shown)


bringdev:
[trello.com/c/J23HFEcB](https://trello.com/c/J23HFEcB)

halvorsanden (approved):

PR #1465, opened 2023-02-13
Convert codetabs to buttons with mb card

halvorsanden:
Wouldn’t have touched this if it hadn’t been in use in the VAS section of service portfolio.
Converts the use of a tags to buttons for toggling between examples and simplifies the JS and replace most of the CSS with mb-card from the design system.


halvorsanden:
Forgot to adjust the dark mode … 💨

Igor-J86 (approved):
LGTM 😃

PR #1467, opened 2023-02-14
Dead links etc

halvorsanden:
Got a report on dead links, so I fixed those and some other accessibility reports:
- Heading levels
- Missing underline on links
- aria-expanded on radio input – it’s already selected, so the expanded isn’t allowed, but it still has the relation to the expanded element


Igor-J86 (approved):
LGTM 😃

PR #1470, opened 2023-02-14
Fix sidemenu ul/li nesting error and reuse template

halvorsanden:
- there was a ul inside a ul without a li element which caused some error reporting
- the subpage navigation was basically a duplication of the main one, it will be easier to modify and clean up when it’s only one place


Igor-J86 (approved):
Really nice! 😃

PR #1471, opened 2023-02-16
updated dangerous goods documentation

Arssul:
- [ ] [API update message](https://github.com/bring/developer-site#tips-for-writing-an-api-update-message) (changelog entry) has been reviewed by Mybring Experience.

image

image

### DIFF
--- a/content/api/booking/examples/vas/DANGEROUS_GOODS.html
+++ b/content/api/booking/examples/vas/DANGEROUS_GOODS.html
@@ -1,10 +1,16 @@
 <div>
   <h3 class="text-heading">Dangerous goods details</h3>
-  <p>When ordering dangerous goods, some additional information describing the goods is required. The following elements/attributes are available:</p>
+  <p>When ordering dangerous goods, some additional information describing the goods is required. The following elements/attributes must be sent if they are defined for the specific UnNumber:</p>
   <ul>
+    <li><code>unNumber</code> (mandatory)</li>
     <li><code>containerId</code> (mandatory when ordered with <code>OIL_EXPRESS</code>)</li>
     <li><code>weightInGrams</code> (either this or)</li>
     <li><code>volumeInLitres</code></li>
-    <li><code>unNumber</code></li>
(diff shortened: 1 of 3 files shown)


content/api/booking/examples/vas/DANGEROUS_GOODS.html
```diff
@@ -1,10 +1,16 @@
...
     <li><code>weightInGrams</code> (either this or)</li>
     <li><code>volumeInLitres</code></li>
-    <li><code>unNumber</code></li>
+    <li><code>commodity</code>  (mandatory if specified)</li>
+    <li><code>packingGroup</code> (mandatory if specified)</li>
+    <li><code>dangerClass</code> (mandatory if specified)</li>
+    <li><code>safetyMarkCode</code> (mandatory if specified)</li>
+    <li><code>tunnelRestrictionCode</code> (mandatory if specified)</li>
   </ul>
+  <p>An example where only unNumber and packingGroup is not enough to verify the correct dangerous good</p>
```
pedromacli, line 15: I'm not sure I understand this sentence? 🤔

pedromacli (approved):
LGTM 👍 (one comment only)

content/api/booking/examples/vas/DANGEROUS_GOODS.html
Arssul: It basically means that for some ADR, by providing just unnr and packingGroup is not enough to find the correct ADR, because there are multiple ADRs that has the same unnr and packingGroup, which is why er want to filter based on tunnelrestriction and then safety mark code as well.

pedromacli: Got it 👍 So I guess we could say "An example where only unNumber and packingGroup **_is provided_** is not enough..."

Arssul: That is better indeed. Fixing it now 👍

PR #1476, opened 2023-02-27
Refactor the dark mode script a bit

Igor-J86:
The script had room for improvements, so code is reduced, but still doing the same. Additionally, I added a function for setting the theme automatically if the mode is set to auto and OS settings change. This will be automatically reflected then without the need of reloading the page.

### DIFF
--- a/css/siteheader.css
+++ b/css/siteheader.css
@@ -119,10 +119,6 @@
   z-index: 1;
 }
 
-.themes-btn {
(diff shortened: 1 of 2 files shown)


bringdev:
[trello.com/c/McqoEC07](https://trello.com/c/McqoEC07)

halvorsanden (approved):
Nice one! 🚢🇮🇹

PR #1484, opened 2023-03-02
Converting the data to json and using OAS template

Igor-J86:
Converting the data source into json and using already established OAS template to render "Send order/article update" for Warehousing.

- Adjusting the template a bit to properly render the data from the source
- Putting all the data into json format
- Converting index.html -> index.md
- Temporary keeping index-old.html


bringdev:

halvorsanden:
I think we should look at the examples in the body schema; they are using XML syntax now, but I think the value is sufficient since everything around it is intended to explain the syntax.

That will also come across clearer by adding the "attribute" property. Things like MessageId and PartyId are XML attributes but are not listed as such in the new one, but it’s listed in the current solution, so we can use that as a reference.

See the Navigation in Address API as an example: https://mango-cliff-0e97d9203-1484.westeurope.azurestaticapps.net/api/address/#get-address-suggestions-get

data/warehousing.json
```diff
@@ -0,0 +1,2731 @@
...
+        ],
+        "type": "string",
+        "properties": {
+          "partyId": {
+            "type": "string",
+            "description": "Customer number (6 or 11 digits) provided by Bring.",
+            "examples": [
+              "<MessageRecipient partyId=20001234567 idType=14 />"
+            ]
+          },
```
halvorsanden, line 469: Ref. the other comment, I think the syntax would be something like:
  
  ```suggestion
            "partyId": {
              "type": "string",
              "description": "Customer number (6 or 11 digits) provided by Bring.",
              "examples": [
                "20001234567"
              ],
              "xml": {
                "attribute": true
              }
            },
  ```

halvorsanden: Used Address as reference: https://api.bring.com/address/api-docs/

data/warehousing.json
```diff
@@ -0,0 +1,2731 @@
...
+        ],
+        "type": "string",
+        "properties": {
+          "subClass": {
+            "type": "string",
+            "description": "Possible values:\n\nDetails on Supplier/Sender/Receiver (required)\n•OrderingParty - Details of Sender\n•DeliveryParty - Details of recipient\n•Supplier - Details of supplier\nFor a PO, the supplier is the one providing the goods to the WH.\nFor a SO, the supplier is you/Bring WH customer.\n(Optional for SO)\n•Buyer - Details of buyer\nOptional\n•Seller - Details of seller.\nOptional\n\nWareHouse ID/Location (required)\n•PerformingUnit\n\nInformation on Transport Company (optional)\n•TransportCompany"
+          },
+          "partyId": {
+            "type": "string",
+            "description": "Party id for the subClass.\n\n\n    Possible values:\n\n    Supplier/Sender/Receiver\n    •provided by user\n\n    Example:\n    <Party subClass=\"OrderingParty\" partyId=\"112233\" />\n\n    Warehouse ID/Location\n    •105075 - Bergen\n    •780132 - Berger/Skedsmokorset\n    •105091 - Drammen\n    •780448 - Fredrikstad\n    •780446 - Haugesund\n    •780429 - KongShavn\n    •105090 - Trollåsen\n    •780135 - Trondheim",
```
halvorsanden, line 1883: Are we able to make these as lists? Doesn’t necessarily have to be done in this PR if there are a lot of them.

data/warehousing.json
Igor-J86: Right.. I could try that instead!

halvorsanden:
XML examples can be added in two ways.
1. If we already have the XML, we can format it as string like https://tracking.bring.com/api-docs/ has done.
2. The other is to do a conversion from the JSON like you have set it up to do but needs some troubleshooting to get it working.

I have no preference.

data/warehousing.json
Igor-J86: Yes, I didn't put to much effort into these descriptions.. Just converted them into json compatible strings, so need to go through these.

halvorsanden: Yup, you decide if that’s something we should do now or later.

data/warehousing.json
```diff
@@ -0,0 +1,3060 @@
...
+      "CreateItemDataRequest": {
+        "required": [
+          "CreateItemDataRequest"
+        ],
+        "type": "object",
+        "properties": {
+          "CreateItemDataRequest": {
+            "$ref": "#/components/schemas/CreateItemData"
+          }
+        }
```
halvorsanden, line 240: The CreateOrderRequest and CreateItemDataRequest objects are only part of XML, not JSON. This is because XML needs a parent object to hold everything.

The extra first level now extra-extra. But can be fixed by essentially replacing the content with the content of the refs (since the refs are only referenced once, they don’t need to be refs)

We don’t show "required" for that level today, so we would have to add support for it. I’m guessing the following can be placed right after the schema object opening:

```json
      "required": [
        "CreateOrderRequest",
        "CreateItemDataRequest"
      ],
```

We have something similar in shipping guide for the `ShippingGuideRequest` object, except the required tag.

data/warehousing.json
```diff
@@ -0,0 +1,3060 @@
...
+        },
+        "description": "Message Details"
+      },
+      "SupplyStructure": {
+        "required": [
+          "ORDERSET"
+        ],
+        "type": "object",
+        "properties": {
+          "ORDERSET": {
```
halvorsanden, line 570: Check the JSON example against these all caps keys. Most of them use camelCase for JSON and only all caps in XML – in which case you need to change the key to the JSON-version and put the XML name in the XML object.

data/warehousing.json
```diff
@@ -0,0 +1,3060 @@
...
+            ],
+            "xml": {
+              "attribute": true
+            }
+          },
+          "updateIndicator": {
+            "type": "string",
+            "description": "Update Indicator",
+            "examples": [
+              "updateIndicator=Deletion"
```
halvorsanden, line 324: For body params, we can just show the value since the syntax varies between XML and JSON.

In contrast to request params that are always in the same format and don’t have full request examples.

data/warehousing.json
```diff
@@ -0,0 +1,3060 @@
...
+            "properties": {
+              "unitCode": {
+                "type": "string",
+                "description": "Unit Code",
+                "xml": {
+                  "attribute": true
+                }
+              },
+              "Value": {
+                "type": "number",
```
halvorsanden, line 1205: Should the number types have "float" as format as well?

Igor-J86: Yes, changing the types to "float" where I can find them.. 😊

data/warehousing.json
Igor-J86: True.. Cleaning up the examples..

data/warehousing.json
Igor-J86: That's right.. Adjusting it..! 😃

data/warehousing.json
halvorsanden: Sorry! Looking at it again, I think most of it is correct the way it is. It’s not camelCase, but PascalCase, as you say. And the JSON examples reflect that. So switching the few all-caps to pascal is enough (except COD, which is all-caps in JSON as well by the looks of the example).

data/warehousing.json
halvorsanden: 👍 and of course there is one that is `int` NumerOfPackages

halvorsanden: Oh, I explained that backwards … 🙈
Those that have `type: number` should have `format: float` (and int for NumberofPackages) in addition, not instead of. SO it ends up something like:
Number <float>

Igor-J86: Oh, actually I've misunderstood here I can see... Will need to adjust that...

PR #1492, opened 2023-03-20
Make XML examples render correctly

halvorsanden:
All objects in the schema were rendered in the XML version of the examples even if they didn’t have any content or weren’t otherwise present in the JSON example.

### before
Screenshot 2023-03-20 at 16 29 40

### after
Screenshot 2023-03-20 at 16 29 24


bringdev:
[trello.com/c/oDeEcHsQ](https://trello.com/c/oDeEcHsQ)

Igor-J86 (approved):

PR #1494, opened 2023-03-22
Prevent service portfolio from freezing (quickfix)

halvorsanden:
A quickfix to prevent service portfolio from freezing in Chrome – have tried a number of things without luck. A working page is better than anything.

Also adjusted the two download links on the same page, they were identical and didn’t register as links in siteimprove.


bringdev:
[trello.com/c/7jJVszgy](https://trello.com/c/7jJVszgy)

PR #1496, opened 2023-03-23
Remove duplication of anchor links in special topics pages

halvorsanden:
- Prevent doubling of anchor links on special topics pages (booking, shipping guide)
- Split sidemenu into to nav elements and remove classes
  - Makes it easier to jump directly to the correct menu, api section and apis without having to trawl through every list element with a screen reader
  - Makes it easier to style with fewer classes and more nested logic (there’s room for doing some actual nesting and similar stuff)
  - Avoids the use of h elements in the nav which triggers a mistake in the structure because the h2 end up before the h1 of the page – by using list elements instead, the hierarchy makes more sense as well.


bringdev:
[trello.com/c/C6pET2M3](https://trello.com/c/C6pET2M3)

halvorsanden:
Here’s the original issue. On the special topics pages "Endpoints" were rendered twice, and the last one plus the SOAP link pointed to the current page, not the parent.
Screenshot 2023-03-23 at 11 47 12

halvorsanden:
Just checked, and the solution doesn’t quite work, SOAP seems to be missing still.

halvorsanden:
Forgot to add a change to the last commit 🤦‍♂️ Should be working now.

halvorsanden:
Should look something like this (+ API support at the end when running in staging/prod):
Screenshot 2023-03-24 at 08 46 19

Igor-J86 (approved):
Nice! 🤩

PR #1505, opened 2023-04-20
Make filtering on API updates page

Igor-J86:
Adding a text input field for filtering on API updates page. This will search through **all** the text of each API update + the API area the update belongs to on `keyup` event. If the input matches the whole, or part of a string, the relevant API updates will be displayed. All others will be hidden.

This could be handled with `includes()` function as well, but using this function things become case-sensitive, so would need to throw some other function on stuff, like `toLowerCase()`, in order to make sure that the input matches parts of the text of each API update. 
I decided therefore to make use of `RegExp` that has a built-in flag for case-insensitivity named `i`, and just `test()` the input value with the API update content and the API area the update belongs to, and then display the relevant API updates.

**Aria-live:**
There are 2 live regions that need to be handled with the `aria-live`, `aria-busy`, `aria-atomic` attributes. If both regions are updated and `aria-busy` is set to `false`, only the last live region was announced with the voice-over tool. Therefore it got a bit trickier to make the logic where the aria-busy is set to false only if there are visible articles in the first region, not the second, and vice versa. 
Setting the `aria-atomic` to `true` is for making sure the whole live region gets announced from start, and not continuing reading out where it left of previously.

**Page:**
https://developer.bring.com/api/revision-history/

Skjermbilde 2023-04-20 kl  13 10 31

### DIFF
--- a/layouts/partials/api/changelog.html
+++ b/layouts/partials/api/changelog.html
@@ -41,27 +41,29 @@
       <label class="form__label" for="filterChangelogInput">Filter API updates</label>
       <input type="text" class="form__control maxw16r" id="filterChangelogInput" />
       {{/* Set $pages to the rearranged structure */}}
-      {{- $pages = $rearrangedPages -}}
-      {{- range first 5 $pages -}}
-        {{- $parentFolder := path.Base (path.Dir .) -}}
-        <article class="mbl" data-changelog-parent="{{$parentFolder}}">
-          <header class="flex flex-dir-col-rev">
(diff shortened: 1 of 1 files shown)


halvorsanden (approved):
Nice use of aria that we should document and set a standard for in similar type of interfaces. I think it’s also time to make a task to look at the layout (again) for this page (maybe we don’t need the collapsible now that we have filtering, or something 🤷‍♂️)
Good work with both the aria and finding a better solution than first assumed.

halvorsanden:
Just ignore the last check missing – working on fixing that.

PR #1506, opened 2023-04-20
Use dashes instead of underscore in job names

halvorsanden:
It’s possible that Terraform has issues when referring to things with underscore – at least the new branch protection doesn’t catch the job name reference.


bringdev:
[trello.com/c/woJ8MmDj](https://trello.com/c/woJ8MmDj)

PR #1518, opened 2023-05-02
Make API updates subscription as web component

Igor-J86:
Making the API updates subscription as a web component with this PR. Not sure if this will work properly with React, so want to test it out by generating a staging static dev-site webapp, and see if the different React apps will allow the web component to be mounted from the served JS file in Mybring.


bringdev:
[trello.com/c/gQUCRUHm](https://trello.com/c/gQUCRUHm)

PR #1520, opened 2023-05-04
Remove SOAP docs again for booking

tomgag:
We want to discontinue SOAP for the Booking API at some point. Let's prevent new users from starting to use it by removing it from the docs.


halvorsanden:
👍 I think you also need to remove the reference to the raml file in config.json (around L100).

tomgag:
> 👍 I think you also need to remove the reference to the raml file in config.json (around L100).

Done

content/api/booking/api.raml
```diff
@@ -2,43 +2,3 @@
 ---
```
halvorsanden, line 2: You can also delete this entire file.

halvorsanden (approved):

PR #1579, opened 2023-08-07
Hide irrelevant parties params in booking API docs

pedromacli:
The `parties` element contains `consignor`, `consignee`, `importer`, `sender`, `recipient` and `returnTo`, which are all implemented by the same DTO ([DtoParty](https://github.com/bring/booking-api/blob/master/src/main/java/no/bring/booking/api/dto/request/DtoParty.java)), so if we add a description for a param in `DtoParty`, said description is shown for all parties. However, not all params/descriptions are relevant for all parties, so we want to hide specific params depending on the party. It doesn't look like it's possible to do this with Swagger (on the `booking-api` side), so doing the changes here. 

@halvorsanden Let me know if there's a better (less hackish) way of achieving this 😛 

Preview of `consignee` party **before:**

consignee before

**After:**

consignee after

### DIFF
--- a/layouts/partials/api/oas/request-schema-xml.html
+++ b/layouts/partials/api/oas/request-schema-xml.html
@@ -1,5 +1,7 @@
 {{- $schemaObj := .schemaObj -}}
 {{- $name := .name -}}
+{{- $objName := .objName -}}
+{{- $contactParent := .contactParent -}}
 {{- $required := .required -}}
 {{- $components := .components -}}
 {{- $endpointId := .endpointId -}}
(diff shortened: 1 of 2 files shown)


bringdev:
[trello.com/c/5Tvq5d9b](https://trello.com/c/5Tvq5d9b)

halvorsanden:
I’ll pull this and try a couple of things …

pedromacli:
Any luck here, @halvorsanden? 😉

halvorsanden:
Yeah, I’m trying to move the exceptions into the API frontmatter so we keep it separate from the template. But haven’t had time to finish it. Since this cannot be done in the source/Swagger (as far as I know too), we have to do something here anyway, so feel free to merge the current solution, and my team can keep working on the other solution and keep you in the loop on that.

halvorsanden (approved):
OK solution for now. We’ll continue to look at another way to do it that doesn’t need API-specific info in the template..

pedromacli:
> Yeah, I’m trying to move the exceptions into the API frontmatter so we keep it separate from the template. But haven’t had time to finish it. Since this cannot be done in the source/Swagger (as far as I know too), we have to do something here anyway, so feel free to merge the current solution, and my team can keep working on the other solution and keep you in the loop on that.

Sounds good :) Thanks for looking into this!

PR #1580, opened 2023-08-07
Add support for topical sub subpages

halvorsanden:
- Rearranges the SG special topics according to their new structure.
- Adds support for a second level and gathering those files in one page.
- Skipped adding a menu on the right-hand side for the individual pages. They didn’t become that long, but we’ll add it to the suggestion box for later.

Screenshot 2023-08-07 at 15 13 53


bringdev:
[trello.com/c/Hthy9xzu](https://trello.com/c/Hthy9xzu)

halvorsanden:
Build fails because event-cast is missing ATM. Seems like it’s being updated.

halvorsanden:
> Looks like some other pages have been affected as well..
> 
> Examples: https://mango-cliff-0e97d9203-1580.westeurope.azurestaticapps.net/api/revision-history/ https://mango-cliff-0e97d9203-1580.westeurope.azurestaticapps.net/api/e-commerce-solutions/

Oooo, funky effect, I only checked the other APIs, thanks!
Maybe it’s time we make some separate templates for such pages.

Igor-J86 (approved):
One small detail, but otherwise this looks good! 🚀

PR #1582, opened 2023-08-10
Prevent loading and linking to language variant of getting started page

halvorsanden:
The page is a category, which in Hugo terms is a collection of the subpages, but we use it as a standalone page (because of historical reasons?), which works fine without hacks, but when one of the pages in that category is translated, the category page also gets the same functionality. The same goes for the home page, but that had a more straightforward check already in place.

Since this came from the checkout guide translation, I discovered a few things and gave them some attention:
- Used the same style for sectioning as the rest of the site (single border instead of complete frame/box)
- Compressed SVGs for web and gave them a reasonable width so they don’t take up too much space when there is no CSS
- Made the responsiveness and structure simpler and better without media queries
- Removed "easy" from the guide descriptions, it didn’t make it easier.


bringdev:
[trello.com/c/mKD6u2SV](https://trello.com/c/mKD6u2SV)

Igor-J86 (approved):
Nice! LGTM 😃

Igor-J86:
Noticed one small bug appearing in the console. Could be handled in this PR. 
Skjermbilde 2023-08-10 kl  09 16 06

PR #1584, opened 2023-08-10
Convert from React to Preact

halvorsanden:
Gives smaller bundle

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,8 +29,8 @@
   },
   "homepage": "https://developer.bring.com",
   "devDependencies": {
-    "@vitejs/plugin-legacy": "^4.1.1",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -50,11 +50,10 @@
     "not ie > 0"
   ],
   "dependencies": {
+    "@preact/preset-vite": "^2.5.0",
```
Igor-J86, line 53: Just a question.. Does it have to be a dependency, or could it be a dev-dependency rather?

halvorsanden: Yup, looks like it can be changed since it’s not Preact in itself, but a help for vite. Good catch!

Igor-J86 (approved):
LGTM 😃

PR #1589, opened 2023-08-23
Replace Mailchimp with CleverReach

Igor-J86:
We are replacing Mailchimp subscription for API updates with the European provider, CleverReach. We had Mailchimp as a web component already, so I have copied this and replaced the template code for it + some adjustments to validation of email format, checkboxes and combination of these.

Still keeping the Mailchimp web component (just in case) in a separate folder with the suffix `-old`, but if things seem to be working out fine, we can completely remove it.


bringdev:
[trello.com/c/ruKlqtFK](https://trello.com/c/ruKlqtFK)

js/webcomponent/api-updates-subscribe-old/README.md
```diff
@@ -0,0 +1,35 @@
+# API updates subscription
+
+A web component made for subscribing to API updates on the developer-site. This
+can be reused whereever it's needed by adding the custom element
```
halvorsanden, line 4: wherever

js/webcomponent/api-updates-subscribe/main.js
```diff
@@ -31,17 +32,20 @@ class ApiSubscribe extends HTMLElement {
...
-        window.$mcj = window.jQuery.noConflict(true)
+      const crValidate = document.createElement("script")
+      crValidate.type = "module"
+      crValidate.src = "https://developer.bring.com/assets/api-updates-subscribe/cr-script.js"
+      this.appendChild(crValidate)
+      crValidate.addEventListener("load", () => {
+        if(typeof jQuery==="undefined"){loadjQuery("//ajax.googleapis.com/ajax/libs/jquery/1.4.4/jquery.min.js",mainCr)}else{mainCr()}
       })
+
+      const reCaptchaScript = document.createElement("script")
```
halvorsanden, line 43: We disabled recaptha right?

js/webcomponent/api-updates-subscribe/template.js
```diff
@@ -1,102 +1,108 @@
...
+
+          <div class="flex flex-dir-col mbs">
+            <fieldset class="flex flex-dir-col">
+              <legend className="mbs">API areas</legend>
+              <div id="8265204" rel="radio" class="cr_ipe_item musthave flex gam" >
+          
                 <label class="form__radio">
-                  <input type="radio" value="All" name="APIAREA" id="mce-APIAREA-0" data-api-area checked>
-                  All
+                    <input id="All8265204" name="1153579" type="radio" value="All" selected checked data-api-area>
```
halvorsanden, line 29: Isn’t _selected_ a `select` attribute? Won’t _checked_ be sufficient?

js/webcomponent/api-updates-subscribe/template.js
```diff
@@ -1,102 +1,108 @@
...
+              <span
+                data-mybicon="mybicon-arrow-right"
+                data-mybicon-width="16"
+                data-mybicon-height="16"
+              ></span>
+              Privacy policy
+            </summary>
+            <div class="text-note pas content__gdprLegal">
+              <p>You can unsubscribe at any time by clicking the link in the footer of our emails. Read more about <a href="https://www.bring.com/privacy-policy" target="_blank" rel="noopener noreferrer">our privacy policy</a>.</p>
+              <p>We use CleverReach. By clicking below to subscribe, you acknowledge that your information will be transferred to CleverReach for processing. <a href="https://www.cleverreach.com/en-de/terms-of-service/" target="_blank">Learn more about CleverReach's privacy practices here.</a></p>
```
halvorsanden, line 90: Pop a couple of `rel="noopener noreferrer"` on the blank links?

layouts/partials/api/changelog.html
```diff
@@ -1,8 +1,6 @@
 <section class="dev-docscontent__section">
   <div class="gaxl flex flex-wrap flex-dir-row-rev justify-cfe">
-    <div class="updates__sub bg-green1 pal align-sfs">
-      <h2>Subscribe</h2>
-      <p class="mbm">Get updates via email or subscribe to our <a href='{{ "api/revision-history/index.xml" | relURL }}'>RSS feed</a>.</p>
+    <div class="updates__sub align-sfs">
```
halvorsanden, line 3: Do we remove the link to the RSS now?

halvorsanden (approved):
Couple of things to look at.

layouts/partials/api/changelog.html
Igor-J86: I think this part could be moved inside the template as well, just below the heading there..?

halvorsanden: Yeah, but wouldn’t that make it appear everywhere the signup form is used?

Igor-J86: That‘s right... Currently the heading and a line of text is placed outside `subscribe.html` partial, and `subscribe.html` is only containing the `<form>`. With the new web component, we could still do the same, but what if it‘s used as a modal?
Should there be a heading and a line of text included then in the template itself, or should it still just be outside of the component? Or if it‘s used a modal, only then include the heading and text?

Igor-J86: Would it be an alternative to place a line of text, including the RSS link, below the component only on the API updates page?

halvorsanden: Yeah, at last have it on the API updates page somewhere. Probably not the most used feature but nice to offer since we have it.

Igor-J86: Agree 😃

PR #1597, opened 2023-09-06
Direct navigation to best practice guides

halvorsanden:
The best practice guides has a starting page for all countries – but the same choices are available in the side menu and on the various country pages. This leads directly to the first one.
Also separated the links to the different versions from the content – and simplified the logic for when to show them.


halvorsanden:
The link in the sidebar in the staging area redirects to prod since the sidebar navigation alias routing looks at the site’s base URL. https://mango-cliff-0e97d9203-1597.westeurope.azurestaticapps.net/api/e-commerce-solutions/best-practice-checkout/norway/ goes directly to the staging area.

cdwinsnes (approved):
🚀

PR #1598, opened 2023-09-06
Add e-commerce feedback form

halvorsanden:
- Moves the checkout CSS into the other CSS bundle
- Adds the form to the e-commerce pages with a button to close it.
- The form slides into view when the footer enters the viewport, but only after 7 seconds (so that it doesn’t pop up right away on the smaller pages). When the user closes the form, it stays hidden for 60 days.

We decided to start simple with the show and hide, and we can adjust it based on feedback and usage.


layouts/partials/checkout/feedback-form-script.html
```diff
@@ -0,0 +1,42 @@
...
+    if (!storedFormDate || storedFormDate < now - delayMs) {
+      const footerTrigger = document.querySelector('footer')
+      const ecomObserverOptions = {
+        root: null,
+        rootMargin: '200px',
+        threshold: 1.0,
+      }
+  
+      function observeHeading(footerTrigger) {
+        for (const footerEl of footerTrigger) {
```
Igor-J86, line 19: Does it need to be a for-loop when it's only one footer element selected?
Also, the function is named `observeHeading`, but should maybe be something like `observeFooter`?

halvorsanden: The loop is needed because the callback doesn’t take the element itself but a list of intersection objects (or changes) on the element. If I understand it correctly, it’s what enables us to have multiple triggers in one element, for instance, counting the percentage visibility of the element. I can try to name it better. I had to read up on it.

It should be footer and not heading, forgot to change the name – I started with the page title and tried to observe when it wasn't visible, but that became too buggy.

Igor-J86: Ah, I see.. Ok.. 😊

halvorsanden: Renamed stuff more to what they are. Seems like "entry" is the term for the change objects: https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API

Igor-J86 (approved):
LGTM! 🚀

PR #1601, opened 2023-09-18
Add XML namespace to booking request examples

pedromacli:
We noticed we were missing `xmlns="http://www.bring.no/booking/"` in the booking API XML request examples, which makes the requests fail. I haven’t found a way of fixing this on the `booking-api` side, so doing the change here. 

**Before:**
xml request before

**After:**
xml request after

@halvorsanden It’s not the first time I’m doing ugly “hacks” in these beautiful templates, so please let us know if there’s a more elegant way of achieving this 😛

### DIFF
--- a/layouts/partials/api/oas/json-to-xml.html
+++ b/layouts/partials/api/oas/json-to-xml.html
@@ -6,6 +6,7 @@
 {{/*  objName doesn’t change  */}}
 {{ $objName := .name }}
 {{ $components := .components }}
+{{ $bookingXmlNamespace := "xmlns=\"http://www.bring.no/booking/\"" }}
 
 {{ $lev := add .lev 1 }}
 {{ $indentChar := "  " }}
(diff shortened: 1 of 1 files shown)


bringdev:
[trello.com/c/F3Fa6vcK](https://trello.com/c/F3Fa6vcK)

halvorsanden:
> 
> @halvorsanden It’s not the first time I’m doing ugly “hacks” in these beautiful templates, so please let us know if there’s a more elegant way of achieving this 😛

🫶 Let me take a look, and if we don’t find a solution in reasonable time, we can just merge and dive deeper for a long term solution.

halvorsanden (approved):
So it doesn’t work because we cannot add xmlns without it also appearing in the JSON-version, right?
I see a couple of other APIs have it, but they have added the entire response as an escaped string, which never uses the converter template. Feels like we should have the option to add these things more directly, but I say merge it for now, and we’ll add it to the backlog along with the other one from last time.

pedromacli:
> So it doesn’t work because we cannot add xmlns without it also appearing in the JSON-version, right?

Yep, that's correct...

Thanks for looking into it 🙂

PR #1603, opened 2023-09-19
Bring back contact details element for recipient party

pedromacli:
...and instead hide the param descriptions for email and phone number, which are only relevant for sender party.

This is a follow-up of [this PR](https://github.com/bring/developer-site/pull/1579).

**Recipient party before (without contact details element visible):**
recipient party before

**Recipient party after (with contact details element visible and hidden param descriptions):**
recipient party after

**Sender party (where we show the contact details element with param descriptions that are relevant for the sender party):**
sender party

### DIFF
--- a/layouts/partials/api/oas/request-schema-xml.html
+++ b/layouts/partials/api/oas/request-schema-xml.html
@@ -68,13 +68,12 @@
     {{- end -}}
   {{- else -}}
     {{/*  Render first level isOf without visible dt  */}}
-    {{/*  Only show contact details element for sender party (booking API docs)  */}}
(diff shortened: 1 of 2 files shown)


bringdev:
[trello.com/c/5Tvq5d9b](https://trello.com/c/5Tvq5d9b)

halvorsanden (approved):
The build fails because there are no more available staging areas. We have a max of 10, which usually is enough, but sometimes old staging areas aren’t deleted when the PR is closed, so they continue to take up space and block new ones even though there are fewer than 10 PRs open.

PR #1606, opened 2023-09-28
Rename E-commerce solutions to Checkout resources

Igor-J86:
Renaming "E-commerce solutions" to "Checkout resources". The url and identifier are still kept as "e-commerce-solutions", but we can later see if this should also be renamed and add alias (redirect) for the old url.

### DIFF
--- a/content/_index.html
+++ b/content/_index.html
@@ -131,7 +131,7 @@ important:
 </div>
 
 <div class="mbxl">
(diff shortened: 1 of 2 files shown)


bringdev:
[trello.com/c/83hGaCEl](https://trello.com/c/83hGaCEl)

halvorsanden (approved):

PR #1618, opened 2023-11-08
Make services_bsg as data source for service icons

Igor-J86:
Adding a fresh `services_bsg.json` where service icons now are added, and making a copy of it to the built folder. 
This will make it publicly available as a data source, so we can make use of it further in Mybring Design System when creating an overview of services and the belonging service icons.


WebDevBring:
[trello.com/c/QA1AsW2m](https://trello.com/c/QA1AsW2m)

halvorsanden (approved):

PR #1625, opened 2023-11-29
Add support for important API updates

Igor-J86:
Adding support for important API updates. By setting `isImportant: true` on the message and including the full API name in the message title, the message will be presented on the right hand side, both on the API page it belongs to, but also on the "API updates" page.

Even though these messages are presented separately from "normal" API updates on the dev-site, they are still part of the regular RSS feed, that is used for distributing the messages by email to the API updates subscribers.

The use case for these are mainly the possibility to flag some messages that should not loose focus for a while, that are of an important nature, and risking of drowning in other API updates that might come in between.

Some examples:
Skjermbilde 2023-11-29 kl  16 45 47

Skjermbilde 2023-11-29 kl  16 53 49

Skjermbilde 2023-11-29 kl  16 55 12

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,16 @@ message is added in each folder (API area) of the affected API.
 - If the message is general and should reach all subscribers, this is added in
 the main folder - `content/api/revision-history`.
 
+- Messages that are important can appear on the API page by setting the parameter
+`isImportant: true` and including the full API name in the message title.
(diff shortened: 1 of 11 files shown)


WebDevBring:
[trello.com/c/7NXmp9Nw](https://trello.com/c/7NXmp9Nw)

PR #1627, opened 2023-12-01
Make the heading appear only when needed

Igor-J86:
Make the "Important updates" heading appear only if there are any "important updates" added and only on the API updates page. The heading was displayed earlier even if no any "important updates" were present.


WebDevBring:
[trello.com/c/7NXmp9Nw](https://trello.com/c/7NXmp9Nw)

PR #1629, opened 2023-12-05
Upgrade dependencies

Igor-J86:


WebDevBring:
[trello.com/c/trCxlgwm](https://trello.com/c/trCxlgwm)

layouts/partials/head.html
```diff
@@ -29,7 +29,7 @@
     href="https://www.mybring.com/design-system/css/main.min.css?ts={{ now.Unix }}"
   />
   {{ if not $.Site.IsServer }}
-    {{- $jsManifest := getJSON "static/assets/manifest.json" -}}
+    {{- $jsManifest := getJSON "static/assets/.vite/manifest.json" -}}
```
halvorsanden, line 32: Any place I can read about this change?

Igor-J86: I cannot find the "why" anywhere, but seems like they just went for it for some reason:
https://github.com/vitejs/vite/pull/14230
https://github.com/vitejs/vite/commit/74fa024

halvorsanden: I see, thanks!

PR #1637, opened 2023-12-12
Turn a long string of utilities into custom class

halvorsanden:
I found out I had done this several months ago but never committed it.

We already had a custom class for it, so I put all the properties in there and used nesting.

Hopefully, we can remove the JS fallback for the has() function next year. Firefox is one or two versions away from supporting it now 🐌


Igor-J86 (approved):
Nice! 🚀

PR #1638, opened 2023-12-12
Removed service 3609 and 3610

LudvigSti:
Removed because there should only be c2c services on the dev site for now

image


WebDevBring:
[trello.com/c/GvAEcGn0](https://trello.com/c/GvAEcGn0)

PR #1646, opened 2024-01-10
Removed mb-disclosure--arrow class

e-kr:
The basic disclosure component is being replaced by the animated arrow disclosure component. 

mb-disclosure--arrow class has therefore been removed as it's being replaced by mb-disclosure. Until https://github.com/bring/mybring-design-system/pull/882 has been merged, it will display two icons in the disclosure. 

Also added mybicon-arrow-right icon to the only disclosure component which didn't use the icon already.


WebDevBring:
[trello.com/c/bks1yahP](https://trello.com/c/bks1yahP)

Igor-J86 (approved):
Nice!

PR #1648, opened 2024-01-19
hide bring pack on dev-site

rijains:
Hide from UI for service "bring pack" from VAS 1288


rijains:
already merged in another PR

PR #1651, opened 2024-01-29
Replace react-instantsearch-dom with react-instantsearch

e-kr:
React-instantsearch-dom has been deprecated, so I have replaced it by react-instantsearch as recommended. It seems to be working exactly as before, on both big and smaller screens, but let me know if you see something I have missed! I also see the completed searches in the analytics page in the algolia dashboard. 

I also removed react-autosuggest as it was not needed anymore. However it turned out that react-instantsearch itself is not really made for the type of dropdown search list, which caused some problems. I read on an algolia forum that it was recommended to use algolia autocomplete along with react-instantsearch in such cases, which is why i ended up using it as well.


WebDevBring:
[trello.com/c/TEi0FZCa](https://trello.com/c/TEi0FZCa)

css/algolia.css
```diff
@@ -1,3 +1,32 @@
...
+  border-bottom-right-radius: 2px;
+  border-bottom-left-radius: 2px;
+  background-color: #fff;
+  box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 1px 5px 0 rgba(0, 0, 0, 0.12);
+}
+
+html[data-theme=dark] .aa-Panel {
+  background-color: var(--green-15);
+  color: var(--green-86);
+}
```
Igor-J86, line 20: This part should be moved inside `dark-theme.css` as a nested class.

css/algolia.css
```diff
@@ -88,6 +88,7 @@ html[data-theme=dark] .react-autosuggest__suggestion--highlighted {
   padding: 0.5rem 0.5rem 0.6rem 1rem;
   border-bottom: 1px solid hsl(0, 0%, 90%);
 }
+
 html[data-theme=dark] .hitlink {
   border-bottom-color: var(--green-darkest);
 }
```
Igor-J86, line 94: Should also go inside dark-theme.css as a nested class.

js/search/AlgoSearch.jsx
```diff
@@ -1,36 +1,57 @@
 import algoliasearch from "algoliasearch/lite"
-import { InstantSearch } from "react-instantsearch-dom"
-import Result from "./Result.jsx"
+import AlgoLogo from "./AlgoLogo.jsx"
+import { InstantSearch } from "react-instantsearch";
+import { Autocomplete } from "./Autocomplete.jsx";
+import React from 'react';
```
Igor-J86, line 5: No need to import React here.

js/search/Autocomplete.jsx
```diff
@@ -0,0 +1,50 @@
+import React, { useEffect, useRef, useState } from "react";
+import { render } from "react-dom";
+
+import { useSearchBox } from "react-instantsearch";
+import { autocomplete } from "@algolia/autocomplete-js";
+import { Fragment } from "react";
```
Igor-J86, line 6: I think the imports can be combined into one line: 

```suggestion
import React,{ useEffect, useRef, useState, render, Fragment } from "react";
import { useSearchBox } from "react-instantsearch";
import { autocomplete } from "@algolia/autocomplete-js";
```

js/search/ResultItem.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
```
Igor-J86, line 1: No need to import React here either.

js/search/AlgoSearch.jsx
```diff
@@ -1,36 +1,57 @@
...
+          },
+          getItems() {
+            return getAlgoliaResults({
+              searchClient,
+              queries: [
+                {
+                  indexName: __ALGOLIA_INDEX__,
+                  query,
+                  params: {
+                    hitsPerPage: 15,
```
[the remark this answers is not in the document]
e-kr: Good suggestion, looks better with 10 😄

js/search/Autocomplete.jsx
```diff
@@ -1,9 +1,7 @@
-import React, { useEffect, useRef, useState } from "react";
+import React, { useEffect, useRef, useState, Fragment } from "react";
 import { render } from "react-dom";
```
Igor-J86, line 2: I think `render` can also be imported from "react"..?

e-kr: yep my bad 😅 will update!

Igor-J86 (approved):
Great job! Seems to be working fine with this approach! 🚀

PR #1653, opened 2024-01-31
removed firefox fallback js for toggle all buttons

e-kr:
Removed firefox fallback js for toggle all buttons, since firefox now supports has(). Tested in latest firefox version (122.0)


WebDevBring:
[trello.com/c/blJApebK](https://trello.com/c/blJApebK)

halvorsanden (approved):

PR #1659, opened 2024-02-12
Use safeHTML function to mark urls separated by <br> tags as safe html.

Oeitho:
- This ensures that the <br> tag is not escaped in the output.


WebDevBring:
[trello.com/c/r58DGdgw](https://trello.com/c/r58DGdgw)

Oeitho:
Relevant documentation on the safeHTML function:
https://gohugo.io/functions/safe/html/

halvorsanden (approved):

PR #1660, opened 2024-02-12
Avoid empty queries algolia search

e-kr:
Updated Algolia search code to avoid sending unnecessary requests when search input is empty


WebDevBring:
[trello.com/c/m0JSBTOb](https://trello.com/c/m0JSBTOb)

halvorsanden (approved):

e-kr:
Update: turns out it was sending duplicate requests because it was sending one request for autocomplete and one request for instantSearch. It should be possible to use both of them together, but I realised that using instantSearch isn't really necessary for the search to work the way we want, and changed to just use autocomplete. Everything looks and works the same, except it doesn't send duplicate requests now

halvorsanden (approved):

PR #1666, opened 2024-03-05
Remove most Express services from docs

tomgag:
As Express IT, the TMS system for Express services is shut down, remove most Express services from the Service overview. An exception being 3336 Express Same Day.

Also removing `service_portfolio.xlsx`, as I couldn't find any references to it.


WebDevBring:
[trello.com/c/i4BrJgO7](https://trello.com/c/i4BrJgO7)

cdwinsnes (approved):

PR #1667, opened 2024-03-08
Patch and minor dependency updates

Igor-J86:
Patch and minor dependency updates. There is also one `hugo-bin` update, but the newer version causes some strange things to happen in conjuction with Vite it seems. I couldn‘t quite figure out of it, so we can handle that in a separate task.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,50 +9,49 @@
       "version": "2.0.0",
       "license": "ISC",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


WebDevBring:
[trello.com/c/yk4FdkuR](https://trello.com/c/yk4FdkuR)

e-kr (approved):

PR #1668, opened 2024-03-13
Update Hugo and replace .CurrentSection condition

Igor-J86:
After last Hugo update it seems like files are served from disk by default and not the memory when running the Hugo server, causing some funkiness with Vite and HMR when spinning up the app locally. Setting the option `--renderToMemory` for Hugo server seems to do the trick in conjuction with Vite.

We also had a conditional for api updates that was checking if the _index files were part of `.CurrentSection` (the output would previously be "revision-history/_index.md").
Seems like the `.CurrentSection` has changed now, so it only gives the parent folder path a.k.a the section of any given file.
Changing the condition to rather check the `.File.Path` works at least, and everything seems to be back to normal.


WebDevBring:
[trello.com/c/WCHy0U2b](https://trello.com/c/WCHy0U2b)

halvorsanden (approved):
Quite the digging job done here 🥇

PR #1671, opened 2024-03-25
feat: add revision about CORS limitations in tracking api

stianste:
These look fine:

image
image

But for some reason, on the tracking API page, it does not show up at the top. Anyone have any idea why this might be? 😕 
image

PR #1672, opened 2024-03-26
Place api updates messages before important ones

halvorsanden:
This makes the updates and important messages more chronological. We want the new update messages to appear first since the important message solution is not for news anymore and its content will be older.

Did some simplification of the spacing and structure between those elements as well.

This leads to changes on the API updates page and the individual API pages, the Booking API has an example of both message types if you want to compare prod and staging.

### DIFF
--- a/css/documentation.css
+++ b/css/documentation.css
@@ -99,7 +99,7 @@ p a:not(.btn-link--dark, .btn-link) {
 
 .dev-docscontent__grid {
   display: grid;
-  gap: 0 2rem;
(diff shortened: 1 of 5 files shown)


WebDevBring:
[trello.com/c/nCtZvrn2](https://trello.com/c/nCtZvrn2)

halvorsanden:
The need was pointed out in https://github.com/bring/developer-site/pull/1671

e-kr (approved):

PR #1681, opened 2024-04-05
Set default typography for documentation

e-kr:
Removed typography css, so that it now uses default typography, and added mb-docs class in order to get the general documentation style added in [this pr](https://github.com/bring/mybring-design-system/pull/925), for documentation specific styling. So this pr will be merged after the pr in design system.

Also found use of the deprecated font size class `text-heading`, so I removed it from h3 since it is now the same size after changing to default styling (1.25 rem), and the places where h2 had the class I turned into h3 instead. The remaining uses I changed to text-1.25r class instead.


css/base.css
```diff
@@ -78,12 +11,6 @@ _:-ms-fullscreen,
   max-width: 106ch;
 }
 
-kbd,
-samp {
-  font-family: monospace, monospace;
-  font-size: 1em;
-}
-
```
halvorsanden, line 86: Yes, these don’t seem to be in use, nice catch!

halvorsanden (approved):
Very nice cleanup of the CSS and the deprecated heading style.

PR #1686, opened 2024-04-11
Edit Event Cast doc

halvorsanden:
- Incorporates Q & A into topic sections, most of which could be categorised as limitations. FAQ/Q&A tend to lead to scattered documentation making it difficult to find topics and cases. 
- Tries to limit the use of package, shipment and parcel for the same thing, at least within the same area.
- Uses more active language to be clearer and shorter.
- Uses "events" instead of both "events" and "event groups", as they seem to be the same thing, but keeps the reference to the event_groups array name of course.

See two comments about events

### DIFF
--- a/content/api/event-cast/index.md
+++ b/content/api/event-cast/index.md
@@ -99,20 +99,18 @@ documentation:
   - title: Payload
     content: |
 
-      ### Payload delivery headers
+      ### Delivery headers
 
-      `HTTP POST` payloads that are delivered to your Webhook's configured URL endpoint contains several Bring specific headers:
+      `HTTP POST` payloads that are delivered to your webhook's configured URL endpoint contains several Bring specific headers:
 
       | Header | Description |
       |:-------|:--------|
       | `X-Bring-Application` | Identifies the application that sent the request to your endpoint |
-      | `X-bring-Correlation` | Correlation Id can be used whilst contacting Bring on error cases |
-      | `X-bring-Version` | Indicates the version of application |
-      | `<your headers>` | Any headers that were specified in the Webhook configuration will also be appended |
+      | `X-bring-Correlation` | Correlation ID can be used when contacting Bring on error cases |
+      | `X-bring-Version` | Indicates the application version |
+      | `<your headers>` | Any headers you specified in the webhook configuration |
 
       ### Example
-
-      Requests issued from Bring to your endpoint will look similar to this:
 
       ```json
       POST /callback/Webhook HTTP/1.1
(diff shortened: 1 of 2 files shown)


WebDevBring:
[trello.com/c/1K3oBzvY](https://trello.com/c/1K3oBzvY)

content/api/event-cast/index.md
```diff
@@ -29,64 +39,54 @@ information:
 documentation:
   - title: Events
     content: |
-      When registering for a Webhook you can choose between a range of different event groups that you can subscribe for.
-      It is recommended to limit the amount of events to subscribe for. Bring will send an event on each of the events occurred which in result may send quite a few HTTP-requests to your server.
-      So, as a good practise, keep the list of event groups as small as possible.
+      We recommend subscribing only to events relevant to you. By keeping the list of events (`event_groups`) as small as possible, you will not get more updates than you need and your server will not get unnecessary HTTP requests from Bring.
+
+      Events are defined as an array with comma separated strings. Wildcards like `*` and `ALL` are not supported.
```
halvorsanden, line 44: What happens if the user doesn’t specify any events? Do they get all or none?

content/api/event-cast/index.md
```diff
@@ -29,64 +39,54 @@ information:
...
       | `PRE_NOTIFIED` | EDI message for the package has been received by Bring. |
       | `READY_FOR_PICKUP` | Package has arrived at pickup point. |
-      | `RETURN` | The package is on its way back to the sender. |
+      | `RETURN` | Package is on its way back to the sender. |
       | `TRANSPORT_TO_RECIPIENT` | Package has been loaded for delivery to the recipient. |
-      | `TERMINAL` | The package is now registered/arrived at inbound/outbound storage terminal |
+      | `TERMINAL` | Package is now registered/arrived at inbound/outbound storage terminal. |
 
-      There is also a specific set of events that will be sent if something deviates from the normal event flow:
+      ### Deviation events
```
halvorsanden, line 76: Are these sent regardless of what the user subscribes to?

content/api/event-cast/index.md
m-ashish: I tried hitting the register endpoint in QA without any events_group and it gave back Bad request with "EventGroup ([]) is either unknown or undefined" message

halvorsanden: OK, thanks, I can add something about that.

content/api/event-cast/index.md
m-ashish: I confirmed with the team, and these events (NOT_REGISTERED and EXPIRED) are not relevant anymore. We can remove this section completely.

content/api/reports/index.md
```diff
@@ -32,7 +32,7 @@ information:
     content: |
       REST XML/JSON over HTTP.
 
-oas: https://qa.mybring.com/reports/v3/api-docs
+oas: https://www.mybring.com/reports/v3/api-docs
```
m-ashish, line 35: Is this relevant with the `event-cast` application? I see you have made the change in `reports`.

halvorsanden: No, not related to event cast, but I had to do it because the qa one prevented the build from finishing. Could have done it in a separate PR though.

m-ashish (approved):
LGTM

PR #1688, opened 2024-04-16
remove b ssh and show from docs

maur1:


WebDevBring:
[trello.com/c/L3oGNig2](https://trello.com/c/L3oGNig2)

deepalimamtani3112 (commented):
Just a small suggestion, on the same page , it also mentions about `ruby` in the line : `Currently there are scripts written in python, ruby and bash.` So Ruby can be removed from eher as no longer used.

deepalimamtani3112 (approved):

sharebear (commented):
This is the source of the blog post published on the public developer site here https://developer.bring.com/blog/b-scripts/

Is there really value in updating 7 year old blog posts? Personally I think we could leave this as it was.

maur1:
> This is the source of the blog post published on the public developer site here https://developer.bring.com/blog/b-scripts/
> 
> Is there really value in updating 7 year old blog posts? Personally I think we could leave this as it was.

Agreed, did not see it in its full context. Closing PR

PR #1689, opened 2024-04-17
Remove 1073 from 9650 could this cause problems dev sit l ke to qnf

EirikStudsrodOlsen:
Updates services and vas list in dev-site with latest from https://github.com/bring/mybring-service-country-vas-common


WebDevBring:
[trello.com/c/lKeToQnf](https://trello.com/c/lKeToQnf)

PR #1690, opened 2024-04-24
Added logging to api-updates-subscribe element, added siteimprove script

e-kr:
Added logging for the api-updates-subscribe custom element when user opens modal and when clicking Subscribe. It only logs when clicking subscribe if inputs are valid. Also added the script from design system.

Check if you agree with the logging values :)

### DIFF
--- a/js/webcomponent/api-updates-subscribe/main.js
+++ b/js/webcomponent/api-updates-subscribe/main.js
@@ -53,7 +53,7 @@ class ApiSubscribe extends HTMLElement {
     // Log to SiteImprove when opening the modal
     if(typeof SiteImprove !== 'undefined'){
       SiteImprove.registerEvent({
-        category: 'Flow steps',
+        category: 'Flow step',
         action: 'Subscribe to API updates',
         label: 'Open subscribe modal'
       })
--- a/js/webcomponent/api-updates-subscribe/main.js
+++ b/js/webcomponent/api-updates-subscribe/main.js
@@ -156,7 +156,7 @@ class ApiSubscribe extends HTMLElement {
         SiteImprove.registerEvent({
           category: 'Task',
           action: 'Subscribe to API updates',
-          label: 'Subscribe to API updates'
+          label: 'Confirm subscription'
         })
       }
     }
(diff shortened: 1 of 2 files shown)


WebDevBring:
[trello.com/c/Cd0BQCmA](https://trello.com/c/Cd0BQCmA)

js/webcomponent/api-updates-subscribe/main.js
```diff
@@ -141,6 +150,15 @@ class ApiSubscribe extends HTMLElement {
           "Invalid email format!"
         )
       }
+
+      // Log to SiteImprove when clicking Subscribe
+      if(typeof SiteImprove !== 'undefined' && (apiAreaRadios[0].checked || someSelected)){
+        SiteImprove.registerEvent({
+          category: 'Task',
+          action: 'Subscribe to API updates',
+          label: 'Subscribe to API updates'
```
Igor-J86, line 159: A suggestion after discussions:

```suggestion
          label: 'Confirm subscription'
```

js/webcomponent/api-updates-subscribe/main.js
```diff
@@ -49,6 +49,15 @@ class ApiSubscribe extends HTMLElement {
     if (closeModalBtn) {
       closeModalBtn.focus()
     }
+
+    // Log to SiteImprove when opening the modal
+    if(typeof SiteImprove !== 'undefined'){
+      SiteImprove.registerEvent({
+        category: 'Flow steps',
```
Igor-J86, line 56: We will update the documentation in MDS so we use singular form on the predefined categories:

```suggestion
        category: 'Flow step',
```

js/webcomponent/api-updates-subscribe/main.js
e-kr: What about "Confirm email subscription" to also include email? Or is it not necessary 🤔

Igor-J86: I think we‘ve concluded that it is sufficient with "Confirm subscription". 😄

Igor-J86 (approved):
Nice! 🚀

PR #1701, opened 2024-05-14
chore: Add message about unauthenticated requests for tracking api

M-Smeby:


WebDevBring:
[trello.com/c/tkvz62Vp](https://trello.com/c/tkvz62Vp)

content/api/tracking/index.md
```diff
@@ -11,6 +11,15 @@ menu:
 weight: 31
  
 important:
+- type: warn
+  title: The Tracking API will no longer support unauthenticated requests from 13.05.2024
+  message: |
+    Starting from May 13th, unauthenticated requests will no longer have access to the Tracking API. Previously, we implemented rate limiting measures.
+    Read here on how to make authenticated requests.
```
richas2, line 18: here should be a link to https://developer.bring.com/api

content/api/revision-history/tracking/2024-05-13.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Unauthenticated Open Tracking API is now deprecated
```
richas2, line 2: Tracking API: Deprecation of Unauthenticated requests

content/api/revision-history/tracking/2024-05-13.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Tracking API: Deprecation of Unauthenticated requests
+publishDate: 2024-05-13
+layout: api
+---
+Starting from May 13th, unauthenticated requests will no longer have access to the Tracking API. Previously, we implemented rate limiting measures.
+Read here on how to make authenticated requests. ``https://developer.bring.com/api/tracking/``
```
richas2, line 7: Shouldn't the here link take directly to `https://developer.bring.com/api` rather than Tracking API page?

M-Smeby: I feel like it is natural for us to link to the page with our API, especially when one of the first things on the page is `Authentication`

content/api/revision-history/checkout/2023-05-15.md
```diff
@@ -3,6 +3,7 @@ title: ShippingGuide API
 publishDate: 2023-05-15
 layout: api
 notanapi: true
+isImportant: true
```
richas2, line 6: Should this be added here?

M-Smeby: Not sure, saw that the message we made about CORS had it, so I thought it was needed. Since I guess this is important 🤔

richas2: But why are we adding this tag in Shipping guide API section? Shouldn't it be in Tracking API section? I am not following this

M-Smeby: Added this to the wrong message 😢 😆

M-Smeby: I somehow ended up in the wrong file, so this was supposed to be added to our message and not this

content/api/revision-history/tracking/2024-05-13.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Tracking API - Deprecation of Unauthenticated requests
+publishDate: 2024-05-13
+layout: api
+isImportant: true
+---
+Starting from May 13th, unauthenticated requests will no longer have access to the Tracking API. Previously, we implemented rate limiting measures.
+Read here on how to make authenticated requests.
```
richas2, line 8: Didnt we use to add the link to the words like this.

`[here](https://developer.bring.com/api)`
image

Like the way we did last time
image

M-Smeby: Yes, that's true

richas2 (approved):
The only thing that I still don't like is that we are forcing user to click multiple times to reach the page where they can learn about how to authenticate an API, not making easy for them. 

Rest looks to good to me now.

M-Smeby:
@richas2 We can change it to use: `https://developer.bring.com/api` but then I think we should do the same in the message from tracking

richas2:
> @richas2 We can change it to use: `https://developer.bring.com/api` but then I think we should do the same in the message from tracking

There we should actually link to this API update message, I don't know if thats possible or not.
But we can replace that link in Tracking error message with this message , that is also fine with me
`Read [here](https://developer.bring.com/api/tracking/) on how to make authenticated requests.`

Issue #1695, opened 2024-05-05
List styling on the blog doesn't look nice anymore


kaaveland:
Seems like `<ul></ul>` would benefit from some more styling on the blog, this doesn't look so nice anymore:

<img width="699" alt="image" src="https://github.com/bring/developer-site/assets/27794/ab903953-40fe-4cb6-a8b0-b34e5fa8b320">

I've actually randomly been dropping onto the blog from google now and then over the past couple of years, which is fun and makes me feel like it's worth it!

halvorsanden:
Oooh, an issue, how nice! 😁
We will fix it, thanks @kaaveland

Igor-J86:
Thanks for the heads up! 👍🏼  A fix is added. ☝🏼

PR #1704, opened 2024-05-21
Add mb-docs class name to blog content

Igor-J86:
Adding the class `mb-docs` to blog content parent `div`, where the same spacing and list styling can be used the same way as for the rest of the site.


WebDevBring:
[trello.com/c/xSeaTfVq](https://trello.com/c/xSeaTfVq)

halvorsanden (approved):
Confirmed fix of https://github.com/bring/developer-site/issues/1695

PR #1705, opened 2024-05-22
Convert custom made modals in web components to `<dialog>`

Igor-J86:
Converting the custom made modals that are developed as web components with this PR. By using the `<dialog>` element, we get some functionality and accessibility for free. Because of that we can reduce some amount of code.


WebDevBring:
[trello.com/c/xPcgyInd](https://trello.com/c/xPcgyInd)

halvorsanden (approved):

PR #1707, opened 2024-05-27
Update event cast documentation

Kohmann:
These changes are result of [YAK cleanup trello card](https://trello.com/c/jpktprO1/3892-verify-info-about-headers-in-dev-site)
### Context
1. It is stated that XML is a supported data format when calling event-cast api, but this is false.
2. The retry service when shipping number is not registered in Bring is also removed.
3. Users have miss-understood the callback example and tried to add `Accept: application/json` as header to requests, which does not work

### Fix:
1. Remove XML as supported format from docs
2. Remove old retry service paragraph 
3. Rename callback example to be more clear that the callback is what Bring sends to the user

### DIFF
--- a/content/api/event-cast/index.md
+++ b/content/api/event-cast/index.md
@@ -83,13 +83,10 @@ documentation:
       ```
       "yyyy-MM-dd'T'HH:mm:ssZ"
       ```
-
-  - title: Payload
-    content: |
-
-      ### Delivery headers
-
-      `HTTP POST` payloads that are delivered to your webhook's configured URL endpoint contains several Bring specific headers:
+      
+      ### Information Bring will provide in the request
+      
+      `HTTP POST` payloads that are delivered to your webhook's configured URL endpoint contains both a JSON with event information and several Bring specific headers:
 
       | Header | Description |
       |:-------|:--------|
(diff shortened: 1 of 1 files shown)


stianste (approved):

content/api/event-cast/index.md
```diff
@@ -98,8 +95,8 @@ documentation:
       | `X-bring-Version` | Indicates the application version |
       | `<your headers>` | Any headers you specified in the webhook configuration |
 
-      ### Example
-
```
richas2, line 102: I think we can just remove `Accept: application/json`, as it says that we expect a Json response from Customer's webhook URL, which is not the case?

richas2 (approved):
One suggestion , rest looks good

content/api/event-cast/index.md
Kohmann: We do expect a json response from them after notifying on their webhook url. It is explicitly stated that we want json in the [WebhookCastService](https://github.com/bring/mybring-events-cast/blob/cd0205a699f9a536353162cd0110b3298eccde0b/src/main/java/com/bring/notifications/webhook/cast/WebhookCastService.java#L55), so we should not remove it.

PR #1709, opened 2024-05-30
add customer subscription docs

Kohmann:
## Launching customer subscriptions to Event cast
Update the Event cast API docs to enable users to create customer number based webhook subscriptions. 
QA is updated with the latest events-cast swagger documentation but the [PR](https://github.com/bring/mybring-events-cast/pull/698) is not merged yet.

### DIFF
--- a/content/api/event-cast/index.md
+++ b/content/api/event-cast/index.md
@@ -48,19 +48,19 @@ documentation:
       Specify the events (`event_groups` / `eventSet`) you want to subscribe to as an array with comma separated strings.
       
       <details>
-      <summary style="cursor: pointer; font-weight: bold;">Tracking number subscription example</summary>
-      
-      ``` 
-      "event_groups": ['IN_TRANSIT', 'NOTIFICATION_SENT', 'TERMINAL']
-      ```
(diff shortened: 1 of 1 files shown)


content/api/event-cast/index.md
```diff
@@ -39,11 +45,25 @@ information:
 documentation:
   - title: Events
     content: |
-      Specify the events (`event_groups`) you want to subscribe to as an array with comma separated strings.
-
-      ```
+      Specify the events (`event_groups` / `eventSet`) you want to subscribe to as an array with comma separated strings.
+      
+      <details>
+      <summary style="cursor: pointer; font-weight: bold;">Tracking number subscription example</summary>
```
stianste, line 51: What is this markup language? Should perhaps `<summary` be indented to show that its a child of `<details>`, or is that not how it works?

stianste (approved):

content/api/event-cast/index.md
Kohmann: Ah yes it is a child of `details` so it should be to more clear with an indent! I fix 
And correct, this is [how you create a collapsable section in Markdown](https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab) :)

PR #1711, opened 2024-06-03
Use api-docs from production to generate event-cast documentation

eideeivindsen:
Looks like [this PR](https://github.com/bring/mybring-events-cast/pull/719) did the trick. The api-docs should be available here now: https://api.bring.com/event-cast/api-docs

### DIFF
--- a/content/api/event-cast/index.md
+++ b/content/api/event-cast/index.md
@@ -143,5 +143,5 @@ documentation:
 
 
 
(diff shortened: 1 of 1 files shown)


WebDevBring:
[trello.com/c/S8aIs9JB](https://trello.com/c/S8aIs9JB)

Kohmann (approved):

PR #1723, opened 2024-06-20
Show OpenAPI Spec link

Igor-J86:
Exposing the OpenAPI Specification link publicly, just below the "Endpoints" table in each OAS based template.

Skjermbilde 2024-06-20 kl  13 01 01


WebDevBring:
[trello.com/c/w0I0Y26G](https://trello.com/c/w0I0Y26G)

halvorsanden (approved):
Looks good, would it be more correct to call it "OpenAPI document"? Ref: https://swagger.io/specification/

Igor-J86:
> Looks good, would it be more correct to call it "OpenAPI document"? Ref: https://swagger.io/specification/

Ye, I think we can go with that! 😄

PR #1731, opened 2024-07-24
Fix dark mode contrast errors

e-kr:
Fixed dark mode contrast errors. Many were caused by having red text on gray background, so I had to add another red color which is a lighter version of red-light, but still darker than red-lighter. Called the variable red-light-lighter, open for better name suggestions 😅

### DIFF
--- a/css/dark-theme.css
+++ b/css/dark-theme.css
@@ -49,6 +49,10 @@
     }
   }
 
+  & .dev-eta1__number, .dev-eta2__number {
+    color: var(--green-light)
+  }
+
   & .dev-docs__menu,
   & .bg-gray1,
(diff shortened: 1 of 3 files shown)


WebDevBring:
[trello.com/c/f7aIcEfH](https://trello.com/c/f7aIcEfH)

halvorsanden (approved):
I think this is fine. I agree that the naming convention is starting to become a bit funky. A separate task probably, but we can consider using the light percentage instead, like we do with greys. So far we’ve used the Bring colour names, and when we make variants outside of that, we use percentage value. Another thing is that if something is used in only one place, making a new variable isn’t necessary.

PR #1732, opened 2024-07-29
use percentage naming convention on color variables

e-kr:
Updated color variable names to use percentage instead of darker lighter etc. 

Considered dropping red-91 and red-93 variables since they are both just used one place, but if we were to need more red colors later they may be useful, and it's easy to forget that those reds exist when they are only specific classes 🤔


WebDevBring:
[trello.com/c/INOHLDoC](https://trello.com/c/INOHLDoC)

halvorsanden (approved):
Several of these are just copies from the design system, so that’s probably why the red ones are there even though they are used only once. I was thinking we could use a reference to the original colour names, but I say we try without for now and add it as comments if we find out it becomes difficult to work on.

PR #1733, opened 2024-07-30
Fix aria label for algolia search input

e-kr:
- Replaced existing aria-label value provided by Algolia by using query selectors, removed label element added previously. Solves Wave orphaned label alert
- The combo-box around the input field got “No accessible text found” in Silktide, and in screen reader it doesn’t indicate what the input is about at all, so I added aria-label to that as well

### DIFF
--- a/js/search/Autocomplete.jsx
+++ b/js/search/Autocomplete.jsx
@@ -11,7 +11,6 @@ export function Autocomplete(props) {
     if (!containerRef.current) {
       return undefined;
     }
-
     const search = autocomplete({
       container: containerRef.current,
       renderer: { createElement, Fragment, render: () => {} },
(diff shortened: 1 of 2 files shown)


WebDevBring:
[trello.com/c/cIAYegSh](https://trello.com/c/cIAYegSh)

js/search/Autocomplete.jsx
```diff
@@ -28,6 +27,15 @@ export function Autocomplete(props) {
       ...props,
     });
 
+    const labelValue = "Search documentation";
+    // change value of aria-label provided by algolia
+    containerRef.current.querySelector(".aa-Label").setAttribute('aria-label', labelValue);
+
+    // removed aria-labelledby provided by algolia which doesn't work
+    containerRef.current.querySelector(".aa-Autocomplete").removeAttribute("aria-labelledby");
```
Igor-J86, line 35: It doesn‘t seem to remove the attribute when testing here, but might be that we don‘t have to remove it if we are setting the `aria-label` as well on the element?

e-kr: hmm it gets removed to me 🤔 but yes it should be ok even if it doesn't get removed, tested locally by not removing it and still works and no wave alerts
image

Igor-J86: I just inspected the Azure staging static site that was generated above here, and the aria-labelledby was still present here, even after refreshing the page couple of times. Not sure why though... 🤔

e-kr: I removed the line and it still works as expected in azure staging site as well, so I think it's not necessary to have it!

Igor-J86 (approved):
LGTM 🚀

PR #1736, opened 2024-08-05
Remove table CSS now served from MDS

halvorsanden:
https://github.com/bring/mybring-design-system/pull/970 serves the same table styles across all of Mybring and documentation.

https://github.com/bring/mybring-design-system/pull/972 will also make the borders of the schemas similar since they use the mb-border class.


WebDevBring:
[trello.com/c/hEaMl6Tu](https://trello.com/c/hEaMl6Tu)

Igor-J86 (approved):

PR #1737, opened 2024-08-05
fix algolia search results without padding

e-kr:
In shipping guide API there are some algolia results which get zero padding, for example it happens if you search Get alternative expected delivery dates, which has Lead Time page as parent. It also happens for sites which has Using Shipping Guide API as parent. 

In order to avoid these pages with zero padding when searching, I am instead linking to the parent component, but adding anchor link so it scrolls down to the correct place. 

Since the anchor tags are made from the titles, I am creating anchor urls from the titles using urlize and adding # infront in the list.algolia.json file. I then add this anchor url to the permalink and relpermalink url of the parent, and then use it for the permalink and relpermalink if the site has these specific parents (Lead Time and Get Alternative Expected Delivery Dates). Using identifiers (sgleadtime and sgusing) of the parent instead of title incase it changes in the future.

Didn’t find any more cases than for pages with Lead Time and pages with Get alternative expected delivery dates as parent.

### DIFF
--- a/layouts/_default/list.algolia.json
+++ b/layouts/_default/list.algolia.json
@@ -44,21 +44,14 @@
 
     {{- $anchorUrlText := .Title -}}
     {{- $anchorUrlText | urlize -}}
-    {{- $anchorUrl := (printf "#%s" $anchorUrlText) -}}
-    {{- $permalink := .Permalink }}
-    {{- $relpermalink := .RelPermalink }}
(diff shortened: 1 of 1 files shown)


WebDevBring:
[trello.com/c/IgmroS2L](https://trello.com/c/IgmroS2L)

Igor-J86 (approved):
LGTM 🚀

PR #1739, opened 2024-08-08
Frontpage api updates set max length and add important badge

e-kr:
Added important badge to API updates on the frontpage. Moved title to be below the date and important badge, as done in API updates page. 

Added a “Show more” button if text length is more than 300 (can change to less). Considered using a link instead but there’s not any links given in the params, so would need to search the text for links or link to Event cast etc based on it having Event cast in the title. I also think it’s nice being able to read the entire update on the frontpage if wanted, since there are links in the posts if relevant and a Show all updates link underneath as well 🤔

### DIFF
--- a/css/frontpage.css
+++ b/css/frontpage.css
@@ -29,3 +29,20 @@
 .start__col {
   flex: 1 1 20rem;
 }
+
+/* Truncate API updates text content on frontpage */
+.full-content {
+  display: none;
+}
+
+.full-content {
+  display: none;
+}
+
+.update-content.expanded .full-content {
+  display: block;
+}
+
+.update-content.expanded .truncated-content {
+  display: none;
+}
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -55,7 +55,17 @@
                     <span class="mb-badge mb-badge--yellow">Important</span>
                   {{- end -}}
                   <h3 class="text-basic">{{ .Title }}</h3>
-                  {{ .Content }}
+                  <div class="update-content">
+                    <div class="truncated-content">
+                      {{ .Content | truncate 300 }}
+                    </div>
+                    <div class="full-content">
+                      {{ .Content }}
+                    </div>
+                    {{- if gt (len .Content) 300 -}}
+                      <button class="read-more btn-link--dark">Show more</button>
+                    {{- end -}}
+                  </div>
                 </div>
               {{- end -}}
               <div class="w100p">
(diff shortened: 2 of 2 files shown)


WebDevBring:
[trello.com/c/x7Dq1vZA](https://trello.com/c/x7Dq1vZA)

layouts/index.html
```diff
@@ -47,13 +47,25 @@ <h2 class="mbs">API updates</h2>
...
+                    <span class="mb-badge mb-badge--yellow">Important</span>
+                  {{- end -}}
+                  <h3 class="text-basic">{{ .Title }}</h3>
+                  <div class="update-content">
+                    <div class="truncated-content">
+                      {{ .Content | truncate 300 }}
+                    </div>
+                    <div class="full-content">
+                      {{ .Content }}
+                    </div>
```
halvorsanden, line 64: Have you considered if or how we can make this accessible to screen readers that shouldn’t have to click "show more"?
We have something similar in the customer service application where I think the full text is always accessible and the truncated text and the button is not. Visually it’s the same though.

layouts/index.html
```diff
@@ -66,4 +78,18 @@ <h3 class="text-basic">
...
   </main>
+
+  <script type="text/javascript">
+    var readMoreLinks = document.querySelectorAll(".read-more");
+    readMoreLinks.forEach(function(link) {
+      link.addEventListener("click", function() {
+        var postContent = link.closest(".update-content");
+        if (postContent) {
+          postContent.classList.toggle("expanded");
+          console.log(postContent)
```
halvorsanden, line 89: Snip off the console.log() before merging to prod. 😸

e-kr: oops 😅

layouts/index.html
e-kr: Good point, probably better for screen reader users to not need to click show more! Going to look into it 😄

css/frontpage.css
```diff
@@ -29,3 +29,20 @@
...
+.full-content {
+  display: none;
+}
+
+.update-content.expanded .full-content {
+  display: block;
+}
+
+.update-content.expanded .truncated-content {
+  display: none;
```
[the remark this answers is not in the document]
e-kr: updated so it uses db / dn utility classes for displaying and hiding content, and only uses the other classes for targetting the elements :)

layouts/index.html
e-kr: Updated so that the truncated text and the button has aria-hidden, and changed the full text to have .screen-reader-text class, so that it is hidden but read by screen readers. And when the truncated text is opened the .screen-reader-text class is removed so its visible and still accessible to screen readers

PR #1741, opened 2024-08-12
revert padding result fix as # in link is causing problem with new search

e-kr:
previous changes to fix results with padding using anchor tag is causing problems, reverting for now


WebDevBring:
[trello.com/c/Rm7Zd5UW](https://trello.com/c/Rm7Zd5UW)

halvorsanden (approved):

PR #1742, opened 2024-08-12
fix issue when using anchor link in algolia search result

e-kr:
Added back changes from this pr https://github.com/bring/developer-site/pull/1737 which i reverted since I thought they caused this problem with search result not appearing. But turned out to happen any time a # was in the url, so the problem already existed. 

Solution ended up being to add panelContainer so the panel container is in the same div as input, and setting panel top to auto. dev-siteheader__search being relative caused problems on smaller screens, so just removed it to make it static. Then the search icon dissapeared so did some changes to make it show without having to use relative.

### DIFF
--- a/css/algolia.css
+++ b/css/algolia.css
@@ -12,6 +12,10 @@
 .algolia-input {
   padding: 0.5rem 1.5rem;
   width: 100%;
+
+  & > div {
+    position: relative;
+  }
 }
 
 .aa-InputWrapperPrefix {
@@ -42,7 +46,7 @@
 .aa-Panel {
   width: 100%;
   color: #222;
-  position: fixed;
+  position: absolute;
   display: block;
   box-sizing: border-box;
   z-index: 10;
@@ -51,14 +55,9 @@
   border-bottom-left-radius: 2px;
   background-color: #fff;
   box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 1px 5px 0 rgba(0, 0, 0, 0.12);
-  top: auto !important;
-}
-
-/* Make results scrollable when smaller screen */
-@media (max-width: 64em) {
-  .aa-Panel {
-    position: absolute;
-  }
+  top: 100% !important; 
+  left: 0 !important; 
+  width: 100% !important;
 }
 
 /* Result item list */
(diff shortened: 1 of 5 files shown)


WebDevBring:
[trello.com/c/Rm7Zd5UW](https://trello.com/c/Rm7Zd5UW)

css/algolia.css
```diff
@@ -52,17 +51,18 @@
   border-bottom-left-radius: 2px;
   background-color: #fff;
   box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 1px 5px 0 rgba(0, 0, 0, 0.12);
+  top: auto !important;
```
Igor-J86, line 54: I tested this by setting `position: absolute;` and removing `top: auto;`. That seems to work..? 
If so, you can also then remove the @media query below here. 😄

e-kr: Did you test on an url with # in it? Like if you search "Get alternative expected delivery dates" and get # in the url or by clicking on an anchor header link

css/algolia.css
```diff
@@ -1,21 +1,20 @@
...
   left: 2rem;
   top: 50%;
   transform: translateY(-50%);
   z-index: 1;
 }
 
 .algolia-input {
-  flex: 1;
+  padding: 0.5rem 1.5rem;
+  width: 100%;
```
Igor-J86, line 14:
```suggestion
  width: 100%;
  
  & > div {
    position: relative;
  }
```

css/algolia.css
Igor-J86: Ah, right. So, I think you can solve it by adding `top: 100% !important; left: 0 !important; width: 100% !important;` and adding `position: relative;` to the div (like I commented above).

css/algolia.css
```diff
@@ -52,17 +51,18 @@
...
 }
 
 /* Make results scrollable when smaller screen */
-@media (max-width: 1000px) {
-  .aa-Panel{
+@media (max-width: 64em) {
+  .aa-Panel {
     position: absolute;
   }
 }
```
Igor-J86, line 62: If that works out, then this is not needed I suppose.

Igor-J86 (approved):
LGTM 🚀

PR #1747, opened 2024-08-26
Remove api from algolia search result and add more info

e-kr:
- Results now have API info if it’s from an API in the label in the search results, so it’s easier to see which API a page belongs to:
image

- Removed “API - “ at start of the label if there is a parent, since most of them include "API" at the end. But keeping it when there is no parent because of the services pages.

- Removed br tag from warehouse page since it showed up in search results

### DIFF
--- a/content/api/warehousing3/index.html
+++ b/content/api/warehousing3/index.html
@@ -12,9 +12,9 @@ weight: 2
 apiArea: Warehousing
 
 introduction:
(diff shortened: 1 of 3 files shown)


WebDevBring:
[trello.com/c/ureMYeuo](https://trello.com/c/ureMYeuo)

halvorsanden (approved):
Nice improvement and iteration!

PR #1752, opened 2024-09-04
Remove unnecessary table classes

halvorsanden:
- Remove unnecessary table classes
- Wrap service portfolio tables in a shadow 🦹 (this won’t have an effect until https://github.com/bring/mybring-design-system/pull/982 is merged)

### DIFF
--- a/content/api/booking-shipment/index.html
+++ b/content/api/booking-shipment/index.html
@@ -32,7 +32,7 @@ notanapi: true
 <section class="dev-docscontent__section">
   <h2 class="dev-anchored">Supported features</h2>
 
-  <table class="apitable">
+  <table>
     <thead>
       <tr>
         <th></th>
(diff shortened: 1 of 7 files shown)


WebDevBring:
[trello.com/c/E5C43AkZ](https://trello.com/c/E5C43AkZ)

Igor-J86 (approved):
Looks good!
The only thing is that the overflow-shadow-thingy is not working quite as expected in dark mode, but that‘s maybe something we can adjust later if needed. 😄

halvorsanden:
> Looks good! The only thing is that the overflow-shadow-thingy is not working quite as expected in dark mode, but that‘s maybe something we can adjust later if needed. 😄

Ooouf, that’s something I should fix now 😬

Igor-J86:
> > Looks good! The only thing is that the overflow-shadow-thingy is not working quite as expected in dark mode, but that‘s maybe something we can adjust later if needed. 😄
> 
> Ooouf, that’s something I should fix now 😬

Ahh, that was nice! Feel free to merge if ready! 👍🏼

PR #1754, opened 2024-09-06
Broken link fixes

e-kr:
- Broken link in info box at Reports API page. Apparently all customers are converted now so the info is outdated and the entire box can be removed
- Updated broken link in booking/authorization page
- Was getting broken link errors in SiteImprove for the Order and Warehousing link at the bottom of Getting Started page, even though they seem to work for me. Removing ./ in the href to see if resolves the issue


WebDevBring:
[trello.com/c/lJj7iZ66](https://trello.com/c/lJj7iZ66)

halvorsanden (approved):

PR #1764, opened 2024-09-24
Move modify delivery to the tracking tile on the front page

arewg:
We decided to move the Modify Delivery API to the tracking tile:

- Tracking team is responsible for the application
- To find Modify Delivery on Mybring, you have to go to Tracking, not Booking. Now it is reflected on the developer site.
- Modifications can only be done to an existing shipment, if you needed to stop and return a shipment while booking it, you would just not book the shipment.
- Although the developer site is publicly available, it is also used internally, and it can confuse the ownership of Modify Delivery
- Having it under Tracking might help guide external questions to the right point of contact.

### DIFF
--- a/content/_index.html
+++ b/content/_index.html
@@ -43,12 +43,11 @@ important:
       <h3 class="mlxs fwn">Booking</h3>
     </div>
     <p class="mhxs pll">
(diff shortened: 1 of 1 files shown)


WebDevBring:
[trello.com/c/7aBe0Ysy](https://trello.com/c/7aBe0Ysy)

stianste (approved):
ITY 🙉

PR #1770, opened 2024-10-04
Split out Pickup API from Booking API

tomgag:
Make Pickup API its own thing.


WebDevBring:
[trello.com/c/VoPkuIfD](https://trello.com/c/VoPkuIfD)

PR #1782, opened 2024-11-13
Developer api doc for document search wl bw jzr8

Mudenz:


WebDevBring:
[trello.com/c/wlBWJzr8](https://trello.com/c/wlBWJzr8)

kiangol:
Looks OK, but I think the Swagger needs some additional information 
- "Search for documents" should have a request body example(s). What is the expected format for the date parameter?
- "Fetch document by id" says 'Returns a bytearray formatted as either JSON or PDF' - how can the content type be set? Example response(s) here would be nice too.

halvorsanden (approved):

Mudenz:
> Looks OK, but I think the Swagger needs some additional information
> 
> * "Search for documents" should have a request body example(s). What is the expected format for the date parameter?
 I will add request body example for the search. 
 
> * "Fetch document by id" says 'Returns a bytearray formatted as either JSON or PDF' - how can the content type be set? Example response(s) here would be nice too.
The files themselves are either JSON or PDF, you can only fetch files in the format they are provided, see the search response. I didnt include any sample responses because these are huge files. Also the API does not interact with the content of the files, so that may change without us knowing.

PR #1784, opened 2024-11-18
Clean up templates

Igor-J86:
Since we don‘t have any pages that are hidden (by the usage of `hidden` param) and there’s no need for it, we are removing the support for `hidden` param.
In addition to that, the JS code for the docs had potential to be improved. I couldn‘t quite see how to improve the examples part further, but there were other parts that could be refactored and reduced, so took the liberty of going through and making some tweaks here and there. 😃


WebDevBring:
[trello.com/c/jMqD9rdN](https://trello.com/c/jMqD9rdN)

PR #1799, opened 2024-12-03
Show deprecated params in api docs

Igor-J86:
Adding support for showing the param state `deprecated: true` (like we have for `required: true`).
The class name in css was quite specific to the `required` state, so making that also a bit more universal.

### DIFF
--- a/css/dark-theme.css
+++ b/css/dark-theme.css
@@ -236,7 +236,7 @@
     }
   }
 
-  & .param-required {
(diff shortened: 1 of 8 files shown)


WebDevBring:
[trello.com/c/hT69tTKX](https://trello.com/c/hT69tTKX)

halvorsanden (approved):

PR #1819, opened 2025-01-28
rename bag on door

rijains:
We need to update the name of the additional service currently referred to as "Pose på døren" in the Mybring customer interface. The service will now be referred to as "Levering utenfor døren".


WebDevBring:
[trello.com/c/EKbx0l23](https://trello.com/c/EKbx0l23)

cdwinsnes (approved):

PR #1823, opened 2025-02-06
Patch and minor dependency updates

Igor-J86:
Some patch and minor updates that were not picked up by dependabot.


WebDevBring:
[trello.com/c/UWRsCNrA](https://trello.com/c/UWRsCNrA)

e-kr (approved):

PR #1832, opened 2025-02-19
Update QR code documentation

karselj:
QR code is now available for services 0331 and 0333 from Norway.

devsite


PraveenPandey01 (approved):
LGTM

PR #1842, opened 2025-03-06
Fix missing required annotation

halvorsanden:
Fixes a bug where the "required" prop was missing from arrays in the json format in request and responses. XML templates did not have the bug.

Shows min length prop. Can probably figure out a better way to show max and min though.

Screenshot 2025-03-06 at 10 42 12


WebDevBring:
[trello.com/c/IK5CvcAY](https://trello.com/c/IK5CvcAY)

tomgag (approved):
LGTM. Approving based on image

PR #1858, opened 2025-03-28
Handle missing ref objects

halvorsanden:
Missing ref objects breaks the XML template and the entire page, the issue must be fixed in the source, but the template shouldn’t break, so that’s prevented now along with a console.error with the name of the missing objects.

And this sent me down a validation rabbit hole, which turns out there could be some improvements. So I fixed a couple that was too difficult to save for later.

Screenshot 2025-03-28 at 14 39 37

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -48,7 +48,7 @@
               {{- range first 3 $pages -}}
                 <div class="bg-gray1 bshadow pal rad-a2px">
                   <time class="mrs text-note"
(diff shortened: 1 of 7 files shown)


WebDevBring:
[trello.com/c/1PrfWr0x](https://trello.com/c/1PrfWr0x)

Igor-J86 (approved):
LGTM 🚀

PR #1860, opened 2025-03-31
Output missing schema warning in both consoles

halvorsanden:
Follow up to https://github.com/bring/developer-site/pull/1858
This adds the check to request and response templates, both XML and JSON. And it sends a warning to the console when building or running along with the name of the API. And it changes the error sent to the browser console into a warning.

There might be a better way to check for this earlier than on each render, but at the same time it allows as much of the interface to render without breaking … might look into it when doing a cleanup of the syntax.

Screenshot 2025-03-31 at 17 50 20


WebDevBring:
[trello.com/c/1PrfWr0x](https://trello.com/c/1PrfWr0x)

Igor-J86 (approved):

PR #1877, opened 2025-05-09
Remove 9650 stuff from examples, free return documentation, QR code even

tomgag:
There are now plans to phase out 9650. Removing some docs related to it, to reduce the number of new integrations on it while we plan its decomission.


WebDevBring:
[trello.com/c/M9vvBCYg](https://trello.com/c/M9vvBCYg)

pedromacli (approved):
👍

PR #1878, opened 2025-05-11
Replace 9300 with 9350 in free return example

tomgag:
Also drop XML


WebDevBring:
[trello.com/c/M9vvBCYg](https://trello.com/c/M9vvBCYg)

sharebear (approved):

PR #1889, opened 2025-06-02
Update customer setup field for revised courier and distribution services

uksolanki:
Depends on https://github.com/bring/mybring-service-country-vas-common/pull/485


EirikStudsrodOlsen (approved):
LGTM

PR #1894, opened 2025-06-06
Fix service portfolio tables

halvorsanden:
The analytics table was completely missing after the renaming a few months ago. 🧐
The filters were also misaligned, probably caused by some updated design system CSS – the margins were to blame, of course. So I used the mb-form class instead.


Igor-J86 (approved):
MLTG 🚀

PR #1911, opened 2025-07-08
Introduce search inputs

halvorsanden:
Doing somewhat the same as in https://github.com/bring/mybring-design-system/pull/1115
Using search type inputs where that makes sense, so that we can use Esc or the browser native cross to clear the input.
Fixing up a couple of things in the Algolia style, the results are different from MDS, so not necessary to force them to be the same.


Igor-J86 (approved):
Nice! 🚀

PR #1923, opened 2025-08-19
Hide 9650 as a service from service portfolio on devsite

rathoreraju:
Hide 9650 as a service from service portfolio on devsite


data/services_bsg.json
```diff
@@ -150,25 +150,6 @@
   "SenderCountries" : "-",
   "Destination" : "-",
   "DomesticAllowed" : "NO"
-}, {
-  "ServiceFamily" : "Parcel Norway domestic",
```
EirikStudsrodOlsen, line 154: Is this text that gets generated when running ServiceVasMappingJsonGenerator in common-lib?
I always forget what on the developer-site that gets generated from common-lib and what we write

rathoreraju: common-lib PR - https://github.com/bring/mybring-service-country-vas-common/pull/511

torstein-a (approved):
👍

PR #1932, opened 2025-09-04
Correcting Finnish pickup point type descriptions

cdwinsnes:
SmartPost type is no longer used after we migrated to the new Posti Pickup point API. Lockers are accessible for everyone (within opening hours)


WebDevBring:
[trello.com/c/30bXNUW4](https://trello.com/c/30bXNUW4)

esschul (approved):

PR #1936, opened 2025-09-10
Prevent checkout updates from showing on all pages

halvorsanden:
Embarrassing typo rendered important checkout update on all API pages.


arewg (approved):
Typos happen! :D Good catch!
Image

